PR #2, opened 2024-01-04
feat: boostraping the monorepo

BLKKKBVSIK:
In this PR:
- Adding Melos and the basic structure of monorepo.


BLKKKBVSIK:
The CI is still missing.
We should probably look around [canonical/firmware-updater](github.com/canonical/firmware-updater) for the CI since it'll be a similar usage.

PR #7, opened 2024-01-12
Set up backend, start with hardware security API

nteodosio:
```
$ cd .../proto
$ protoc --go_out=. --go_opt=paths=source_relative --go-grpc_out=. --go-grpc_opt=paths=source_relative     ./dsc.proto
$ go build .../backend/*
# systemd-socket-activate  -l 127.0.0.1:8076 <binary>
```
Other terminal
```
$ curl 127.0.0.1:8076
```
will create /tmp/c.txt.


packages/backend/hardware-sec.go
```diff
@@ -0,0 +1,26 @@
...
+    "google.golang.org/grpc/status"
+    pb "github.com/canonical/desktop-security-center/packages/proto"
+    "google.golang.org/protobuf/types/known/emptypb"
+)
+
+func (s *HardwareServer) IsSecureBootEnabled(ctx context.Context, p *emptypb.Empty) (*pb.IsSecureBootEnabledResponse, error) {
+    cmd := exec.Command("mokutil", "--sb-state")
+    out, err := cmd.Output()
+    if err != nil {
+        return &pb.IsSecureBootEnabledResponse{ Enabled: true }, status.Errorf(codes.NotFound, "couldn't execute mokutil")
```
nteodosio, line 16: Must this really be this painful? I was expecting I could just `return true, status.Errorf(...)`

packages/backend/hardware-sec.go
```diff
@@ -0,0 +1,26 @@
...
+    "strings"
+    "os/exec"
+    "context"
+    "google.golang.org/grpc/codes"
+    "google.golang.org/grpc/status"
+    pb "github.com/canonical/desktop-security-center/packages/proto"
+    "google.golang.org/protobuf/types/known/emptypb"
+)
+
+func (s *HardwareServer) IsSecureBootEnabled(ctx context.Context, p *emptypb.Empty) (*pb.IsSecureBootEnabledResponse, error) {
```
nteodosio, line 12: Is there is no doing without the second argument?

matthew-hagemann: When you compile your protos, you should get output a file named `dsc_grpc.pb.go` which will have interface definitions in it. What you are doing in this file is implementing the `HardwareServer` Interface on your `HardwareServer` struct so that it can be used by your gRPC server

Go interface implementation are [implicit](https://go.dev/tour/methods/10), so in Go you just need to match the method signatures defined for the trait in that generated file to implement the interface, which includes the empty in this case.

What you can do is swap the  `p *emptypb.Empty` to `_  *emptypb.Empty` as you wont use it for anything.

packages/backend/hardware-sec.go
matthew-hagemann: Usually in Go, you will either return a response or an error, and `nil` the other. Because this is an error case, you can do this:
```
if err != nil {
    return nil, status.Errorf(codes.NotFound, "couldn't execute mokutil")
}
``` 
You can do the same for that final else clause that has an error.

On a side note, you could probably swap this from a series of if else statements to a [switch statement](https://go.dev/tour/flowcontrol/9), as all the cases seem distinct to me.

packages/systemd/server.socket
```diff
@@ -0,0 +1,5 @@
+[Socket]
+ListenStream=127.0.0.1:8076
```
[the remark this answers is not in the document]
matthew-hagemann: I am assuming this is going to work on over a unix socket similar to `provd` eventually? If thats the case, when you bring in the [daemon](https://github.com/matthew-hagemann/ubuntu-desktop-provision/tree/main/provd/internal/daemon) modules and set up the [cmd](https://github.com/matthew-hagemann/ubuntu-desktop-provision/tree/main/provd/cmd/provd) files, you can use a testing config that points to a continent address for testing and use that config when testing.

Its quite a lot of code, but it should be fairly easy to bring it in with minimal change to get it up and running. I basically copied it over from one of the other Go projects. Its the infrastructure for spinning up a gRPC server as a deamon listening on a unix socket. 

But for now, whatever helps you test getting the server logic working makes sense to me :slightly_smiling_face:

matthew-hagemann (commented):
Nice work Nathan! You are on the right track! I have replied to all your questions inline. I'm happy to answer more, either in this format or we can hop on a call if you'd like, I'm easy :slightly_smiling_face:

packages/backend/hardware-sec.go
nteodosio: I see, I'll update it to return nil in those cases then, thanks!

But for the other cases I was just annoyed by the verbosity of `&pb.IsSecureBootEnabledResponse{ Enabled: true }` instead of just `true`, but well that was just style in the end of the day.

PR #15, opened 2024-01-25
fix: missing Serve statement for GRPC server

nteodosio:


matthew-hagemann (approved):
Should solve the BE not serving :slightly_smiling_face: 

I am aware this is just code to test your services with, so don't worry about changing it now, but definitely something to keep in mind when you refactor later is error handling / more graceful exiting, not panicking.

Ideally, your new func would return an `error`, and all your panics would change to some kind of log followed by returning an `error`, and return `nil` if you have no issues :slightly_smiling_face:

tim-hm (approved):

PR #16, opened 2024-01-26
Start implementation of the back-end Ubuntu Pro service

nteodosio:


packages/backend/ubuntu-pro.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+func (s *ProServer) IsMachineProAttached(ctx context.Context, _ *emptypb.Empty) (*pb.Boolean, error) {
+    cmd := exec.Command("pro", "api", "u.pro.status.is_attached.v1")
+    out, err := cmd.Output()
+    if err != nil {
+        return nil, status.Errorf(codes.NotFound, "couldn't query Pro status")
+    }
+    val := gjson.Get(string(out), "data.attributes.is_attached")
+    return &pb.Boolean{ Ret: val.Bool() }, nil
```
nteodosio, line 19: This is returning gibberish. In my [simple test client](https://people.canonical.com/~npt/client.go), the call returns

```
&{{{} [] [] %!t(*impl.MessageInfo=&{0x871500 0xc0000e39d0 <nil> [] {0 0} 1 {map[1:0xc000296280] map[] map[] [<nil> 0xc000296280] [0xc000296280] 0x601520 0x601480 0x60cf40 {<nil>}} {{{} 3 0x612100 0x612020 0x612180 0x612320 0x6122a0} [0xc000298240] [<nil> 0xc000298240] map[1:0xc000298240] 8 16 false 18446744073709551615 false false 0}})} %!t(int32=0) [] false}
```

How is this a boolean?!

matthew-hagemann: I don't think the response will be just a bool, it will probably have some wrapping, so trying to print it with `%t` is what I think causing the issue:

You could try printing the full response struct and its field to get a better read of what is happening:

```
    u, err := p.IsMachineProAttached(context, new(emptypb.Empty))
	if err != nil {
		fmt.Println(err)
	} else {
        fmt.Printf(":%+v:%t\n", u, true)
    }
```

Alternatively,  you could use [gRPCurl](https://github.com/fullstorydev/grpcurl) to test, which will do the formatting for you, and you don't need to extend it each time you add a new endpoint you want to test:

```
grpcurl -plaintext -unix -proto locale.proto -d '{"locale": ""}' /home/matt/test.sock locale.LocaleService/SetLocale
```
In the above example:
`locale.proto` = path to the protofile that defines the endpoint you want to test
`'{"locale":""}'` = your request 
`/home/matt/test.sock` = path to your unix socket
`locale.LocaleService/SetLocale` = `proto package name` . `service name` / `rpc name` (all as defined in your protofile)

nteodosio: Thank you for your help, Matt!

I hadn't realized that the result would be in some wrapping.

If I try to print it as suggested, it prints:

```
::true
```

I'll try the tool you suggested and report back.

matthew-hagemann: Great! Glad that helped :smile:

nteodosio: For the record, `grpcurl` agrees with client.go: it is returning `{}`.

nteodosio: OK, if I read https://stackoverflow.com/questions/75352422/grpc-not-returning-boolean-if-value-is-false correctly this is by design. Unbelievable!

nteodosio:
Note for self: We can after all use DBus for all this stuff, no need to exec `pro api`.

I should therefore rewrite the RPC calls that use the latter to use the former. Therefore I'll leave the wrong result of a Is*Enabled query after a {Enable,Disable}* unattended for now.

packages/backend/ubuntu-pro.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func (s *ProServer) DisableInfra (ctx context.Context, _ *emptypb.Empty) (*emptypb.Empty, error) {
+    return new(emptypb.Empty), enableService("esm_2dinfra", "Disable")
+}
+
+func pollProInitiate(reqId string) {
+    cmd := exec.Command("pro", "api", "u.pro.attach.magic.wait.v1", "--args", "magic_token=" + reqId)
+    out, err := cmd.Output()
+    if err != nil {
```
nteodosio, line 124: The `pro api <wait>` command might possibly fail because of 

- Failure exec'ing per se, e.g. inexistent command or insufficient permissions etc. (less likely).
- Failure of communication to the server, e.g. machine lost connection to Internet, in case it returns a number of descriptive error codes (more likely).

This first err is catching both above scenarios, defeating the usefulness of the error code, which would be caught by the next `if` check.

Any suggestion as to the best way of addressing this?

packages/backend/ubuntu-pro.go
```diff
@@ -0,0 +1,163 @@
...
+    return &pb.Boolean{ Ret: enabled }, err
+}
+
+func (s *ProServer) IsKernelLivePatchEnabled(ctx context.Context, _ *emptypb.Empty) (*pb.Boolean, error) {
+    enabled, err := isServiceEnabled("livepatch")
+    return &pb.Boolean{ Ret: enabled }, err
+}
+
+var dbusServices map[dbus.ObjectPath]map[string]map[string]dbus.Variant
+func connectToSystemBus() error {
```
matthew-hagemann, line 56: Currently only the `ubuntu-pro` service makes use of `dbus`, but if you add any services in the future that also make use of `dbus`, I would move this call to connect to the system bus into the server.go file, get the connection once, and then pass it into each of the services.

Best way to do something like that is to define a new function on each service that takes in the `dbus conn` and returns a service. Something like this:
[Dbus connection obtained once](https://github.com/matthew-hagemann/ubuntu-desktop-provision/blob/main/provd/internal/services/manager.go#L30-L33)
[New function takes it in and makes it accessible to the service](https://github.com/matthew-hagemann/ubuntu-desktop-provision/blob/main/provd/internal/services/user/user.go#L44-L76)

Not immediately applicable but something to be mindful of :smile:

packages/backend/ubuntu-pro.go
```diff
@@ -0,0 +1,163 @@
...
+    "os/exec"
+    "context"
+    "errors"
+    pb "github.com/canonical/desktop-security-center/packages/proto"
+    "google.golang.org/protobuf/types/known/emptypb"
+    "github.com/tidwall/gjson"
+    "github.com/godbus/dbus/v5"
+)
+
+var conn *dbus.Conn
```
matthew-hagemann, line 12: On a similar note to the comment about using a new function, I would look at defining a struct that holds all your service variables such as this dbus connection. Even if you are not currently passing in a dbus conn to the service, I would define a new func that does the setup of establishing the connection and then holding it for use by functions in this service

For example:
[Defining a struct that holds different variables relevant to the serivce](https://github.com/matthew-hagemann/ubuntu-desktop-provision/blob/main/provd/internal/services/user/user.go#L30-L37)
[New function that initializes all the variables.](https://github.com/matthew-hagemann/ubuntu-desktop-provision/blob/main/provd/internal/services/user/user.go#L44-L76)
I would also use it to hold dbus objects you, see next comment!

packages/backend/ubuntu-pro.go
```diff
@@ -0,0 +1,163 @@
...
+
+var dbusServices map[dbus.ObjectPath]map[string]map[string]dbus.Variant
+func connectToSystemBus() error {
+    var err error
+    conn, err = dbus.ConnectSystemBus()
+    return err
+}
+
+func enableService(basename string, able string) error {
+    obj := conn.Object(
```
matthew-hagemann, line 63: I would move this to being held in the service struct and initialized in the service new function. 
If you need to reuse dbus objects (which you do here and in `attach`), it makes it easier, but the other advantage is you can ping them in the new function and fail early if your service can't reach the object for whatever reason, instead of only failing when the function is called (conn.Object does not check the object exists, only holds reference it where to find it)
[Initializing and pinging dbus object in the new function](https://github.com/matthew-hagemann/ubuntu-desktop-provision/blob/main/provd/internal/services/user/user.go#L49-L55)

packages/backend/ubuntu-pro.go
```diff
@@ -0,0 +1,163 @@
...
+        dbus.ObjectPath("/com/canonical/UbuntuAdvantage/Services/" + basename),
+    )
+    status, err := obj.GetProperty("com.canonical.UbuntuAdvantage.Service.Status")
+    if err != nil {
+        return false, err
+    }
+    return status.Value().(string) == "enabled", nil
+}
+
+func (s *ProServer) IsMachineProAttached(ctx context.Context, _ *emptypb.Empty) (*pb.Boolean, error) {
```
matthew-hagemann, line 27: A general comment that will apply to all your public functions:
We make use of golangci as a linter. One of the big things it complains about is adding comments to your public functions to make them easier to reason about by consuming modules 
[Example of a .golangci.yaml file](https://github.com/matthew-hagemann/ubuntu-desktop-provision/blob/main/provd/.golangci.yaml)
[golangci linter](https://github.com/golangci/golangci-lint)

packages/backend/ubuntu-pro.go
```diff
@@ -0,0 +1,163 @@
...
+        dbus.FlagAllowInteractiveAuthorization,
+    )
+    if call.Err != nil {
+        return call.Err
+    }
+    return nil
+}
+
+func (s *ProServer) EnableKernelLivePatch(ctx context.Context, _ *emptypb.Empty) (*emptypb.Empty, error) {
+    return new(emptypb.Empty), enableService("livepatch", "Enable")
```
matthew-hagemann, line 78: It generally more idiomatic to return nil as your response if an error has occurred. So I would rather suggest:

```
...
err := enableService("livepatch", "Enable") 
if err != nil {
    // Option to add logging if needed
    return nil, err
}
return new(emptypb.Empty), nil
...
```
Applied to all calls like this one

packages/backend/ubuntu-pro.go
matthew-hagemann: Not something I've had to deal with to much, but if I look at [how the enterprise team is handling similar calls](https://github.com/canonical/ubuntu-pro-for-wsl/blob/main/wsl-pro-service/internal/system/pro.go#L10-L27) it looks like they are including the context and just giving a single error along with the context and allowing that to suffice for multiple error cases :slightly_smiling_face:

matthew-hagemann (changes requested):
Nice work @nteodosio!

I've left some comments, but most of my changes are style / code organization related and I'd rather point them out now than later :slightly_smiling_face:

One other question I have is if you given thought to how you want to approach testing your services? This is also probably the time to think about that!

I've recently finished doing the test harness for `provd`. Dbus is especially a fun one to work with :sweat_smile: but you can end up with tests that look like [this](https://github.com/matthew-hagemann/ubuntu-desktop-provision/blob/main/provd/internal/services/locale/locale_test.go) using some of the [pre-existing testutils](https://github.com/matthew-hagemann/ubuntu-desktop-provision/tree/main/provd/internal/testutils) from the enterprise team and me.

Take a look and let me know. Didier took me through it all initially, and most of it is his design, so he is definitely the domain expert, but it is also now fresh in my head and I'd be happy to show you what I have done for testing `provd` :slightly_smiling_face:

packages/backend/ubuntu-pro.go
matthew-hagemann: The endgame for the linter will ultimately be to have it [run in the ci](https://github.com/matthew-hagemann/ubuntu-desktop-provision/blob/main/.github/workflows/ci.yml#L152-L167) using the desktop team h/desktop-engineering/gh-actions/go/code-sanitycheck. But you can do that another time in another PR

packages/backend/ubuntu-pro.go
nteodosio: Grpcurl always gives me this error if I try it to return nil where an Empty is expected:

```
ERROR:
  Code: Internal
  Message: grpc: error while marshaling: proto: Marshal called with nil
```

matthew-hagemann: Try using [grpc status's](https://grpc.github.io/grpc/core/md_doc_statuscodes.html), I think that will solve your issue:

```
import (
	"google.golang.org/grpc/codes"
	"google.golang.org/grpc/status"
)

...

if err != nil {
	return nil, status.Errorf(codes.Internal, "failed to preform this task: %v", err)
}
```

packages/backend/ubuntu-pro.go
nteodosio: I dealt with it in another way, but it is more robust now, since it also checks whether the response is indeed a JSON.

packages/backend/ubuntu-pro.go
nteodosio: Matt, I think I've addressed your remaining comments now. I'm wondering if that's not more generalizable, could you please have a look?

matthew-hagemann: @nteodosio Hi Nathan. Sure, I'll try take a look later today :slightly_smiling_face:

nteodosio: Thanks. About my concert, I think maybe it is fine, I just need to grow Manager to accommodate more services (and rename functions accordingly) as in, if I wanted to add a permissionsServer:

```
type Manager struct {
    proServer        *ProServer
    permissionServer *PermissionServer
    conn             *dbus.Conn
}

func NewServerManager (ctx context.Context) (*Manager, error) {
    conn, err := dbus.ConnectSystemBus()
    if err != nil {
        return nil, status.Errorf(codes.Internal, "Failed to connect to system bus: %s", err)
    }

    proServer, err := NewProServer(conn)
    if err != nil {
        return nil, status.Errorf(codes.Internal, "Failed to create user service: %s", err)
    }

    permissionServer, err := NewPermissionServer()
    if err != nil {
        return nil, status.Errorf(codes.Internal, "Failed to create permission server: %s", err)
    }

    return &Manager{
        proServer:        proServer,
        permissionServer: permissionServer,
        conn:             conn,
    }, nil
}
```

packages/backend/server.go
```diff
@@ -3,29 +3,68 @@ import (
...
+}
+
+func NewProServerManager(ctx context.Context) (*Manager, error) {
+    conn, err := dbus.ConnectSystemBus()
     if err != nil {
-        panic(err)
+        return nil, status.Errorf(codes.Internal, "Failed to connect to system bus: %s", err)
     }
 
+    userService, err := NewProServer(conn)
```
matthew-hagemann, line 27: I think `userService` is one of mine? :smile:

packages/backend/ubuntu-pro.go
```diff
@@ -0,0 +1,256 @@
...
+        return false, err
+    }
+    return status.Value().(string) == "enabled", nil
+}
+
+/* Determines if system is attached to Ubuntu Pro. */
+func (s *ProServer) IsMachineProAttached(ctx context.Context, _ *epb.Empty) (*wpb.BoolValue, error) {
+    isAttached, err := s.manager.GetProperty("com.canonical.UbuntuAdvantage.Manager.Attached")
+    if err != nil {
+        log.Println(err)
```
matthew-hagemann, line 83: Nothing wrong with go's log package, but we prefer [log/slog](https://pkg.go.dev/log/slog), as it allows for richer log messaging. You don't have to fix it now, you can look into this for another PR, just want to let you know!

Allows you to differentiate between:
```
 slog.Error(fmt.Sprintf("somthing went wrong: %v", err))
 slog.Info("somthing happened!")
 slog.Warn("somthing concerning happened!")
 slog.Debug("somthing a dev might be interested in happened!")
```
The gotcha of slog is that it doesn't currently support formatting strings out the box, so you need to wrap errors in a `fmt.Sprintf(...)`, but you do end up with much nicer logs:

```
2024/02/09 09:41:34 INFO Using configuration file: example.config.yaml
2024/02/09 09:41:35 INFO Serving GRPC requests on /home/matt/test.sock
```

packages/backend/ubuntu-pro.go
```diff
@@ -0,0 +1,256 @@
...
+    err := enableService(s.infra, "Disable")
+    if err != nil {
+        return nil, status.Errorf(codes.Internal, "%v", err)
+    }
+    return new(epb.Empty), nil
+}
+
+/* Waits until the user enters his PIN, retrieved via InitiateProMagicFlow, in
+ * ubuntu.com/pro/attach (or whatever happens to be the appropriate URL at the
+ * time of reading), retrieves and returns the attachment token. */
```
matthew-hagemann, line 185: :tada:

matthew-hagemann (approved):
Awesome work! 

2 minor nitpicks, but I won't block you on them  (you can definitely look into logging in a future PR)

WRT testing, if you'd like, we can hop on a call some time next week and I can show you how I'm testing `provd`? It's enough of an effort to set up that you could make it a different PR, but I wouldn't get too many services in before getting it set up :slightly_smiling_face:

packages/backend/ubuntu-pro.go
nteodosio: Ack! I will address this in a later merge request then.

nteodosio:
Thank you once again, Matt, for your very instructive and detailed review!

matthew-hagemann:
> Thank you once again, Matt, for your very instructive and detailed review!

My pleasure! :slightly_smiling_face:

PR #25, opened 2024-02-19
Test backend: Ubuntu Pro

nteodosio:


packages/backend/ubuntu-pro_test.go
```diff
@@ -0,0 +1,460 @@
...
+        },
+    }
+    for _, tc := range cases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+            execMocker := func(name string, arg ...string) *exec.Cmd {
+                ret := fmt.Sprintf(waitJson, tc.contract, tc.errors, tc.result)
+                return exec.Command("printf", "%s", ret)
+            }
```
matthew-hagemann, line 99: Nice work mocking the exec! What would make this a little cleaner would be to:

- Create the `execMocker` as a type
- Have boolean flags you could set depending on what error you want it to return (ie: connectivity-error)
- Create the mock with the flags and move the logic of what is returned into the mock, not passing it in the test

The main goal would be to make your test cases as concise as possible, so instead of:
```

           name: "Invalid response, has failure",
            wantError: true,
            result: "failure",
            errors: `{
                        "msg": "Failed to connect to authentication server",
                        "code": "connectivity-error",
                        "meta": {}
                     }`,
```

you could have:
```
            name: "Invalid response, has failure",
            wantError: true,
            networkError: true
```

packages/backend/ubuntu-pro_test.go
```diff
@@ -0,0 +1,460 @@
...
+		t.Fatal("I expected no error, but didn't get one")
+	}
+	i, err = manager.proServer.IsEsmAppsEnabled(ctx, nil)
+	if err == nil {
+		t.Fatal("I expected no error, but didn't get one")
+	}
+	i, err = manager.proServer.IsKernelLivePatchEnabled(ctx, nil)
+	if err == nil {
+		t.Fatal("I expected no error, but didn't get one")
+	}
```
matthew-hagemann, line 321: Putting the comment here, but applies to all the if err... checks, I would swap them all for requires as mentioned above :smile:

matthew-hagemann (commented):
Really nice work! I've left some comments all related to styling that I think could help. Might also be worth getting Didier's eyes on this at some point to get his perspective. Having him help you refine this will give you a nice base to work from for future services :slightly_smiling_face:

.github/workflows/ci.yaml
```diff
@@ -28,6 +28,7 @@ jobs:
           protoc --go_out=. --go_opt=paths=source_relative --go-grpc_out=. --go-grpc_opt=paths=source_relative     ./dsc.proto
           cd "$basedir/backend"
           go build .
+          go test
```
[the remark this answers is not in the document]
nteodosio: Can I leave this one for the next pull request? A couple of RPCs were missing and I'd like to switch to that before working on this part.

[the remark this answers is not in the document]
nteodosio: Test reply

matthew-hagemann (approved):
Nice work! I know it's difficult to balance feature development with code/test cleanliness :smile: You could always merge what you have here to unblock yourself and Enzo and circle back later. I have no issues with how its functioning!

I don't want to block you on advisory things, so I will point out anything major that needs changing, but otherwise point things out as advice/tips to work in going forward / when you have time to circle back :slightly_smiling_face:

packages/backend/ubuntu-pro_test.go
nteodosio: Something like this in the last commit?

nteodosio: I'm reverting that last commit because I want to merge this but not with a "surprise" commit. I will include it in the other merge request with the existing Github actions you mentioned in another comment.

PR #26, opened 2024-02-23
Implement the snap permissions back-end

nteodosio:


nteodosio:
As communicated in that thread, authorized calls to that endpoint will be implemented. While this doesn't happen, we will use Polkit.

packages/backend/permissions.go
```diff
@@ -0,0 +1,210 @@
...
+    resBody, err := io.ReadAll(res.Body)
+    if err != nil {
+        log.Printf("Couldn't read response from %s request to %s.", kind, where)
+        return "", err
+    }
+    return string(resBody), nil
+}
+
+func validateApiResponse(res string) error {
+    if !gjson.Valid(res) {
```
sminez, line 69: @matthew-hagemann I had a chat about this on a call with @nteodosio earlier: can you provide some references of how the JSON parsing works in `provd` for @nteodosio to take a look at please? :slightly_smiling_face:

packages/backend/server.go
```diff
@@ -37,9 +40,19 @@ func NewServerManager (ctx context.Context, conns ...*dbus.Conn) (*Manager, erro
         return nil, status.Errorf(codes.Internal, "Failed to create user service: %s", err)
     }
 
+    c := &http.Client{
+        Transport: &http.Transport{
+            DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
+                return (&net.Dialer{}).DialContext(ctx, "unix", "/run/snapd.socket")
```
sminez, line 46: This is something I hit with the prompting client when I was getting it packaged up as a snap: there's a different socket we need to connect to in that case. [This](https://github.com/canonical/aa-prompting-test/blob/main/aa-prompt-client/src/snapd_client.rs#L116-L120) was the simplest way to check for which socket we should use that I could come up with (the `snapd` folks also suggested that this was the way to go).

Not a blocker for merging but something we'll need to remember to take a look at before we package as a snap.

sminez (approved):
As discussed on our call earlier today this should be good to merge as is (the comments are fine to be addressed in a follow up PR :slightly_smiling_face:) but once that is done we'll be looking at iterating on a Flutter UI with embedded snapd client for the current "move fast and change things" work so that we have fewer moving parts to keep in sync. Once that has stabilised we'll catch up with you and go through what changes (if any) need to be made on the backend side.

packages/backend/permissions.go
matthew-hagemann: Sure thing!

@nteodosio  So here is an example from the service that communicated with `pro`

[Here](https://github.com/canonical/ubuntu-desktop-provision/blob/main/provd/internal/services/pro/pro.go#L54-L75) I define the structure of the json response the pro client sends back as a go struct with annotations.

And then you can quite easily [unmarshal](https://github.com/canonical/ubuntu-desktop-provision/blob/main/provd/internal/services/pro/pro.go#L175-L176)  the response and handle any errors, leaving you with a go struct you can work with.

nteodosio: Not fun when the value can be either an array or a 'simple type'.

nteodosio: E.g. the same end-point might return

```
{
  "result": {
    "message": "rule ID is not found"
  }
}
```
or
```
{
  "result": [
    {
      "id": "C7JHBW7E7Q7PO===",
      "timestamp": "2024-05-24T13:48:17.723465463Z",
      "user": 1000,
      "outcome": "allow",
      ...
    }
  ]
}
```

PR #31, opened 2024-02-28
fix: do not error if Is*Service is called while unattached.

nteodosio:
The front-end may call Is{Service}Enabled even if the machine is detached and this shouldn't cause an error. So make isServiceEnabled first check whether Ubuntu Pro is attached before actually trying to call the respective Dbus object.

### DIFF
--- a/packages/backend/ubuntu-pro.go
+++ b/packages/backend/ubuntu-pro.go
@@ -90,13 +90,13 @@ func NewProServer(conn *dbus.Conn) (*ProServer, error) {
     return s, nil
 }
 
-func isServiceEnabled(obj dbus.BusObject, manager dbus.BusObject) (bool, error) {
+func (s *ProServer) isServiceEnabled(obj dbus.BusObject) (bool, error) {
     /* Although it is pointless to call isServiceEnabled if the machine is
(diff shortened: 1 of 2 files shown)


packages/backend/ubuntu-pro.go
```diff
@@ -99,20 +112,27 @@ func isServiceEnabled(obj dbus.BusObject) (bool, error) {
     return status.Value().(string) == "enabled", nil
 }
 
+func isMachineProAttached(manager dbus.BusObject) (bool, error) {
```
matthew-hagemann, line 115: I would change this to have the pro service as a pointer receiver so you don't have to pass the manager around, and maybe rename it to something like `isAttached` to make to a bit clearer this is just a check and fairly different to `IsMachineProAttached`.

Otherwise I think this makes perfect sense to double check the FE usage of these endpoints :smile:

matthew-hagemann (approved):
Makes sense to me! One small change request :slightly_smiling_face: otherwise LGTM!

packages/backend/ubuntu-pro.go
nteodosio: Oh of course that's so much better, refreshing even. Thanks!

PR #37, opened 2024-04-13
feat: add German translation

Atalanttore:
I was unable to determine the exact meaning of “Magic attach” in the "proDialogMagicAttach" string. That is why I have not translated this string.


packages/desktop_security_center/lib/src/l10n/app_de.arb
```diff
@@ -0,0 +1,37 @@
...
+    "ubuntuPro": "Ubuntu Pro",
+    "ubuntuProEnabled": "Ubuntu Pro ist aktiviert",
+    "ubuntuProDisabled": "Nicht geschützt mit Ubuntu Pro",
+    "getUbuntuPro": "Ubuntu Pro erhalten",
+    "attachThisMachine": "Diese Maschine anbinden",
+    "ubuntuProFreeDesc": "Immer kostenlos für den persönlichen Gebrauch.\n30-Tage-Testversion für Unternehmen.",
+    "esmLabel": "Erweiterte Sicherheitswartung (ESM)",
+    "esmDesc": "ESM bietet zusätzliche Sicherheitspatches für 10 Jahre für mehr als 23.000 Open-Source-Pakete. Es ermöglicht ein kontinuierliches Schwachstellenmanagement für kritische, hohe und mittlere CVEs.",
+    "proKernelLivepatchLabel": "Kernel-Livepatch",
+    "proEsmMainPackages": "Hauptpakete (esm-infra)",
```
nteodosio, line 12: As main is the name of the repository, I'd not translate that word and write it as you did in the line below (Main-Pakete).

packages/desktop_security_center/lib/src/l10n/app_de.arb
```diff
@@ -0,0 +1,37 @@
...
+    "proComplianceAndHardening": "Compliance und Härtung",
+    "proComplianceAndHardeningWarning": "Nur zur Einhaltung von FedRAMP, HIPAA und anderen Compliance- und Härtungsanforderungen empfohlen.",
+    "proUsg": "Ubuntu-Sicherheitsleitfaden (USG)",
+    "proUsgDesc": "Automatisiert das Härten und Auditieren mit CIS-Benchmark- und DISA-STIG-Profilen und ermöglicht gleichzeitig umgebungsspezifische Anpassungen. ",
+    "proUsgDocumentation": "USG-Dokumentation",
+    "proFips": "FIPS 140-2",
+    "proFipsDesc": "Ein von der US-amerikanischen und kanadischen Regierung zertifiziertes kryptografisches Modul, das die Einhaltung des Datenschutzstandards FIPS 140-2 bestätigt. ",
+    "proFipsDocumentation": "FIPS-Dokumentation",
+    "proDisablePro": "Ubuntu Pro deaktivieren",
+    "proDialogMagicAttach": "Magic attach with this code at ",
```
nteodosio, line 27: Yes this "magic" is not very useful jargon coming from the Ubuntu Pro API. I'd suggest you to simply ignore that word for the translation.

nteodosio (commented):
Thank you, this looks nice, I've left a couple of comments.

packages/desktop_security_center/lib/src/l10n/app_de.arb
Atalanttore: Even without “magic” I don't really understand the meaning of the "proDialogMagicAttach" string, which would make my translation a game of chance.

packages/desktop_security_center/lib/src/l10n/app_de.arb
Atalanttore: Yes, it would be better not to translate the name of the repository. I have changed it.

packages/desktop_security_center/lib/src/l10n/app_de.arb
nteodosio: Makes sense!

nteodosio (approved):

nteodosio:
Thank you!

PR #42, opened 2024-06-07
Fix CI

nteodosio:


matthew-hagemann (approved):
LGTM! 

Not really needed here because you only have one proto file, but if you find more get added it can be nice to move the compiler step to a [generate.go](https://github.com/canonical/ubuntu-desktop-provision/blob/main/provd/generate.go) file, then you can all all of them using `go generate`. Not needed here, but a cool thing to keep in mind if this grows :slightly_smiling_face:

nteodosio:
Thanks for the suggestion, I incorporated it into this merge request. Do you want to have another look or can I go ahead and merge?

matthew-hagemann (approved):
Awesome :tada:

PR #47, opened 2024-08-22
chore: preparing for publishing the stable build

sminez:
We are currently missing a number of files and config we should have in place for publishing the stable build, such as a README, license and contribution guidelines. We also need to update the desktop entry to use the new icon and the `snapcraft.yaml` to allow us to prompt the revsion to `stable`.


d-loose (approved):
I'll follow up with a PR that sets up melos and fvm later on.
Setting up and using a launchpad bug tracker makes sense to me - we might need to think about disabling the github issue tracker entirely.

PR #48, opened 2024-08-23
feat: use long-polling to update the UI based on snapd's state

d-loose:
Automatically updates the UI when
* the prompting feature is toggled externally
* prompting rules change

This also includes some retry logic to gracefully handle snapd errors while it's restarting when the feature flag is toggled

[Screencast from 2024-08-23 13-58-16.webm](https://github.com/user-attachments/assets/fcb97f90-b346-40e4-9691-b86e76b526c3)

~~The UI change is a bit delayed when the feature is toggled externally, I'll see if I can add a loading state or at least disable the toggle button for that case.~~

Added a `waitingForSnapd` state

[Screencast from 2024-08-23 14-31-18.webm](https://github.com/user-attachments/assets/809b3d01-a20e-4b22-a601-d59d86045410)

UDENG-3790
UDENG-4077

### DIFF
--- a/lib/services/fake_app_permissions_service.dart
+++ b/lib/services/fake_app_permissions_service.dart
@@ -48,11 +48,6 @@ class FakeAppPermissionsService implements AppPermissionsService {
   }
 
   @override
-  Future<void> dispose() async {
-    await _statusController.close();
-  }
-
-  @override
   Future<void> init() async {
     _statusController = StreamController<AppPermissionsServiceStatus>.broadcast(
       onListen: () async {
@@ -60,6 +55,11 @@ class FakeAppPermissionsService implements AppPermissionsService {
         _statusController.add(AppPermissionsServiceStatus.enabled(rules));
       },
     );
+  }
+
+  @override
+  Future<void> dispose() async {
+    await _statusController.close();
   }
 
   @override
--- a/lib/services/fake_app_permissions_service.dart
+++ b/lib/services/fake_app_permissions_service.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
@@ -27,14 +28,7 @@ class FakeAppPermissionsService implements AppPermissionsService {
   }
 
   final List<SnapdRule> rules;
-  bool _enabled = true;
-
-  @override
-  Future<List<SnapdRule>> getRules({
-    String? snap,
-    String? interface,
-  }) async =>
-      rules;
+  late final StreamController<AppPermissionsServiceStatus> _statusController;
 
   @override
   Future<void> removeRule(String id) async {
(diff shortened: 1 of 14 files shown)


lib/services/fake_app_permissions_service.dart
```diff
@@ -54,19 +48,34 @@ class FakeAppPermissionsService implements AppPermissionsService {
   }
 
   @override
-  Future<bool> isEnabled() async => _enabled;
+  Future<void> dispose() async {
+    await _statusController.close();
+  }
```
spydon, line 53: Very nitpicky, but I would place this anywhere underneath `init`

lib/services/fake_app_permissions_service.dart
```diff
@@ -54,19 +48,34 @@ class FakeAppPermissionsService implements AppPermissionsService {
...
+
+  @override
+  Stream<AppPermissionsServiceStatus> get status => _statusController.stream;
 
   @override
-  Stream<AppPermissionsServiceStatus> enable() async* {
-    yield AppPermissionsServiceStatusEnabling(0.0);
+  Future<void> disable() async {
+    _statusController.add(AppPermissionsServiceStatus.disabling(0.0));
     await Future.delayed(const Duration(seconds: 3));
```
spydon, line 71: Is this delay necessary?

lib/services/fake_app_permissions_service.dart
```diff
@@ -54,19 +48,34 @@ class FakeAppPermissionsService implements AppPermissionsService {
...
-    _enabled = true;
+    _statusController.add(AppPermissionsServiceStatus.disabled());
   }
 
   @override
-  Stream<AppPermissionsServiceStatus> disable() async* {
-    yield AppPermissionsServiceStatusDisabling(0.0);
+  Future<void> enable() async {
+    _statusController.add(AppPermissionsServiceStatus.enabling(0.0));
     await Future.delayed(const Duration(seconds: 3));
```
spydon, line 78: Same here

matthew-hagemann:
I found that if you cancel out of the admin password prompt when toggling prompting, the option remains off until you restart the security center:

[Screencast from 2024-08-27 13-42-07.webm](https://github.com/user-attachments/assets/03f2839a-6a41-4b9a-983b-917964e1060b)

d-loose:
> I found that if you cancel out of the sudo password prompt when toggling prompting, the option remains off until you restart the security center:

Very good catch, thanks a lot! I'll need to handle that case explicitly

pubspec.yaml
```diff
@@ -23,7 +23,7 @@ dependencies:
   path: ^1.9.0
   riverpod: ^2.5.1
   riverpod_annotation: ^2.3.5
-  snapd: ^0.6.2
+  snapd: 0.6.4
```
spydon, line 26: How come we need to pin this?

spydon (approved):
Good stuff! The only comments I have are small nitpicks.

lib/services/fake_app_permissions_service.dart
d-loose: In a sense yes - the fake service exists mainly to test the UI, so there's an artificial delay to simulate the loading state with enough time to observe the UI state

pubspec.yaml
d-loose: We don't, I just forgot the caret when replacing my test branch with the new version number :)

matthew-hagemann (approved):
LGTM! This is really awesome for managing the prompting client and the aa rules :slightly_smiling_face: nice work!

PR #49, opened 2024-08-28
feat: use melos and fvm 

d-loose:
This sets up package management with melos and Flutter SDK management with FVM.

P.S.: I've also updated the versions of `prompting-client` and `snapd` in the integration test, since the needed versions are now available in `latest/stable` and `latest/edge`, respectively.

UDENG-4303


CONTRIBUTING.md
```diff
@@ -97,49 +97,57 @@ to minimize diff review.
 
 ### Required dependencies
 
-#### Flutter
-
-[Install Flutter](https://flutter.dev/docs/get-started/install/linux) - the
-currently used version is specified in `.tool-versions`. If you're using
-[asdf](https://asdf-vm.com/) to manage your Flutter SDK, you can simply run
-`asdf install` to install the required version.
+[Install Flutter](https://flutter.dev/docs/get-started/install/linux) - the currently used version is specified in `.fvmrc`. If you're using [fvm](https://fvm.app) to manage your Flutter SDK, you can simply run `fvm install` to install the required version.
```
spydon, line 100: Should we keep 80 (or 100 maybe?) max line length like before, so that it's easier to read in non-wrapping editors

spydon (approved):
That is so much cleaner! Especially the snapcraft file.
Just one tiny comment about markdown line length.

.github/workflows/ci.yaml
```diff
@@ -8,74 +8,94 @@ on:
...
+          restore-keys: |
+            ${{ runner.os }}-repo-
+      - uses: Atsumi3/actions-setup-fvm@0.0.3
+      - uses: bluefireteam/melos-action@v2
+      - name: Install dependencies
+        run: sudo apt update && sudo apt install -y clang cmake libglib2.0-dev libgtk-3-dev liblzma-dev ninja-build pkg-config polkitd xvfb
+      - name: Setup apparmor prompting
+        run: |
+          sudo snap refresh snapd --channel=latest/edge
+          sudo snap install prompting-client
```
sminez, line 95: do we potentially want to do a matrix here over the edge and stable channels of the prompting client?

sminez: The client itself isn't being used directly (I don't think) but if there are any future changes like the ones in snapd that require the client to have certain interfaces present it might help catch any strange interactions that we might otherwise miss.

sminez (dismissed):

.github/workflows/ci.yaml
d-loose: right, the only reason it's being installed is so that we can actually enable the feature, which now requires a client to be present. does it make sense to run those checks here in the CI for the security center, though?

sminez (approved):

PR #68, opened 2024-09-16
feat: check if prompting feature is supported

d-loose:
Checks whether the currently installed snapd version supports apparmor prompting and disables the toggle button if that's not the case.

Also: I've removed the hardcoded log level, so it can be changed via the environment and added some useful debug logs.

UDENG-4493

### DIFF
--- a/packages/security_center/lib/main.dart
+++ b/packages/security_center/lib/main.dart
@@ -15,7 +15,7 @@ import 'package:yaru/yaru.dart';
 
 Future<void> main(List<String> args) async {
   await YaruWindowTitleBar.ensureInitialized();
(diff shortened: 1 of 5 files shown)


d-loose:
Snap is building [here](https://github.com/canonical/desktop-security-center/actions/runs/10885279464), would appreciate if someone could double-check that everything works as intended by installing the snap and testing the following things:
* toggling the prompting feature on/off using snapd 2.65
* downgrading to snapd 2.63 and verifying that the UI doesn't hang in a loading state but instead shows the disabled toggle button

sminez (approved):
@d-loose after getting my testing VM sorted out I can confirm that your fix is working correctly for me :+1:

PR #69, opened 2024-09-17
fix: fix long-polling with unsupported prompting

d-loose:
On older versions of snapd long-polling the notices endpoint with an unknown notice type filter causes the request to return immediately instead of blocking until the timeout passes, which results in constant requests being sent to snapd. To fix this, we simply filter for warning notices in the long-polling loop when talking to a snapd version that doesn't support prompting. This should suffice to detect snapd errors or restarts that warrant a status update in the UI.

UDENG-4493


sminez (approved):
Not been able to test manually but the change looks good 👍

PR #72, opened 2024-09-18
fix: Fixing issue with libhandy on non-GNOME x11

sminez:
Modifying the way we set up the GTK window using libhandy to match how things are done in the app-center. The commit that set things up in the app-center can be found here:
  https://github.com/ubuntu/app-center/commit/43e7ae0a6ce82706b77587899ad06c36a2fac781

Note that this is _not_ how [the docs for the handy_window package](https://pub.dev/packages/handy_window#usage) tell you to set things up (this is what we followed originally). Prior to this change we tested the current code under multiple desktop environments on Ubuntu 24.04 with the following results:

  - GNOME classic Wayland: works
  - GNOME classic X11: works
  - Ubuntu desktop Wayland: works
  - Ubuntu desktop X11: works
  - Penrose X11: uninitialised window, closing leave the process running and eating CPU.
  - i3: uninitialised window, closing leave the process running and eating CPU.

With the change all of the above desktop environments work as expected.

> The above are tested on an AMD Framework 13

---

:warning: **We will need to make an equivalent change to the firmware updater as well** :warning: 

---

@robert-ancell: Dennis and I spent the morning getting to the bottom of this and while we are confident that this PR fixes the issue we're not 100% on what the underlying problem is, particularly as the existing code _does_ work under GNOME on X11. Any ideas?

### DIFF
--- a/packages/security_center/linux/my_application.cc
+++ b/packages/security_center/linux/my_application.cc
@@ -1,9 +1,7 @@
 #include "my_application.h"
 
 #include <flutter_linux/flutter_linux.h>
-#ifdef GDK_WINDOWING_X11
-#include <gdk/gdkx.h>
-#endif
+#include <handy.h>
 
 #include "flutter/generated_plugin_registrant.h"
 
@@ -17,35 +15,9 @@ G_DEFINE_TYPE(MyApplication, my_application, GTK_TYPE_APPLICATION)
 // Implements GApplication::activate.
 static void my_application_activate(GApplication* application) {
   MyApplication* self = MY_APPLICATION(application);
-  GtkWindow* window =
-      GTK_WINDOW(gtk_application_window_new(GTK_APPLICATION(application)));
 
-  // Use a header bar when running in GNOME as this is the common style used
-  // by applications and is the setup most users will be using (e.g. Ubuntu
-  // desktop).
-  // If running on X and not using GNOME then just use a traditional title bar
-  // in case the window manager does more exotic layout, e.g. tiling.
-  // If running on Wayland assume the header bar will work (may need changing
-  // if future cases occur).
-  gboolean use_header_bar = TRUE;
-#ifdef GDK_WINDOWING_X11
-  GdkScreen* screen = gtk_window_get_screen(window);
-  if (GDK_IS_X11_SCREEN(screen)) {
-    const gchar* wm_name = gdk_x11_screen_get_window_manager_name(screen);
-    if (g_strcmp0(wm_name, "GNOME Shell") != 0) {
-      use_header_bar = FALSE;
-    }
-  }
-#endif
-  if (use_header_bar) {
-    GtkHeaderBar* header_bar = GTK_HEADER_BAR(gtk_header_bar_new());
-    gtk_widget_show(GTK_WIDGET(header_bar));
-    gtk_header_bar_set_title(header_bar, "security_center");
-    gtk_header_bar_set_show_close_button(header_bar, TRUE);
-    gtk_window_set_titlebar(window, GTK_WIDGET(header_bar));
-  } else {
-    gtk_window_set_title(window, "security_center");
-  }
+  GtkWindow* window = GTK_WINDOW(hdy_application_window_new());
+  gtk_window_set_application(window, GTK_APPLICATION(application));
 
   GdkGeometry geometry;
   geometry.min_width = 800 + 52;
@@ -53,16 +25,16 @@ static void my_application_activate(GApplication* application) {
   gtk_window_set_geometry_hints(window, nullptr, &geometry, GDK_HINT_MIN_SIZE);
 
   gtk_window_set_default_size(window, 800+52, 600+52);
+  gtk_widget_show(GTK_WIDGET(window));
 
   g_autoptr(FlDartProject) project = fl_dart_project_new();
   fl_dart_project_set_dart_entrypoint_arguments(project, self->dart_entrypoint_arguments);
 
   FlView* view = fl_view_new(project);
+  gtk_widget_show(GTK_WIDGET(view));
   gtk_container_add(GTK_CONTAINER(window), GTK_WIDGET(view));
 
   fl_register_plugins(FL_PLUGIN_REGISTRY(view));
-  gtk_widget_show(GTK_WIDGET(window));
-  gtk_widget_show(GTK_WIDGET(view));
 
   gtk_widget_grab_focus(GTK_WIDGET(view));
 }


d-loose (dismissed):
I think that's the cleanest solution for now :+1:

spydon (dismissed):
Lgtm, does the general flutter template need some of these changes?

sminez:
> Lgtm, does the general flutter template need some of these changes?

The issue is that we need to modify the general template when we add `handy_window`. I don't _think_ the changes are needed / wanted without it? But I'm not 100% sure.

One thing I didn't call out above is that setting up a new hello world app using flutter 3.19.6 and running it straight away without adding any packages works just fine on my system. We pinned things down to working before `handy_window` was added and breaking with the commit that added it. So it's definitely an issue on that side rather than a general one.

robert-ancell (changes requested):
This looks a lot simpler than using the handy_window plugin. Looking at the plugin, it modifies the class which is pretty complicated. I'm not sure if that's considered safe, and I suspect on some systems it's not overriding everything that is used.

You should remove the use of handy_window with this change, and modify CMakeLists.txt to depend on libhandy.

I don't think we can upstream this change like this though, as it would make a default dependency on libhandy, which doesn't seem an essential library.

sminez:
> You should remove the use of handy_window with this change, and modify CMakeLists.txt to depend on libhandy.

@robert-ancell We can do but that's not the setup in `app-center` so I'd like to pin down what the "correct" way of doing this is and then use that consistently everywhere. Are you saying that we can just remove the dart `handy_window` package entirely and that the change to `my_application.cc` along with updating `CMakeLists` is all that we need? If so I really don't understand what the dart package is actually doing / why it is a dart package...

> I don't think we can upstream this change like this though, as it would make a default dependency on libhandy, which doesn't seem an essential library.

:+1: good to have that confirmed. From my POV its an opinionated change that _we_ want for Ubuntu applications but not one that makes sense as the default for everyone.

d-loose (approved):
I stand corrected - this is a much cleaner solution! Thanks @robert-ancell for pointing this out

sminez:
@d-loose and I took a look at this again this morning and https://github.com/canonical/desktop-security-center/pull/72/commits/f86cc9b501e90f68290fcaf54e80f5ba77d37778 contains the changes needed to build and run with libhandy directly instead of using the `handy_window` dart package. If we are happy with this approach we should make the equivalent change in our other flutter projects and look into what we want to do about the `handy_window` package as it looks like it's not the way this should be handled going forward?

spydon (commented):
What are the actual differences here over using handy_window? Seems like shame to have to depend on a native package that we already have bundled in a flutter package. Maybe these changes could be made in that package instead?

d-loose:
Hm, an interesting quirk I just noticed is that the rounded corners disappear when activating high contrast mode in the accessibility settings. We should double-check that the we don't use any essential functionality by manually adding `libhandy` instead of using `handy_window`.

sminez:
Fun (type 2), the saga continues :sweat_smile: 

I was mentioning this issue to @local-optimum and he was pointing out that he still doesn't have rounded corners on the 24.10 daily build at the moment (but that he _does_ on 24.04). So it sounds like we need to try and pin down exactly what set up does / doesn't work in the cases we're able to test for and ideally cover all of our bases :crossed_fingers: 

CI on that last push is broken (I didn't update the deps in the workflow to include libhandy) so I'll fix that now, but are we saying we _don't_ think this is OK to merge in its current state?

d-loose:
> CI on that last push is broken (I didn't update the deps in the workflow to include libhandy) so I'll fix that now, but are we saying we _don't_ think this is OK to merge in its current state?

I'd wait for @robert-ancell's approval. I've created a snap build [here](https://github.com/canonical/desktop-security-center/actions/runs/10941288995) to do some more testing in the meantime.

robert-ancell (approved):
Changes look good for changing GtkWindow for HdyWindow. However, as discussed is there anything else the handy_window package is doing? It does seem to have some [settings code](https://github.com/ubuntu/handy_window.dart/tree/main/linux/settings) which I'm not sure why it is needed.

You might want to check with the handy_window authors if there's anything you'll lose.

Feichtmeier:
> Changes look good for changing GtkWindow for HdyWindow. However, as discussed is there anything else the handy_window package is doing? It does seem to have some [settings code](https://github.com/ubuntu/handy_window.dart/tree/main/linux/settings) which I'm not sure why it is needed.
> 
> You might want to check with the handy_window authors if there's anything you'll lose.

The only other minimal thing that it is doing is applying a style to the window shadow that is similar to gtk4 windows, however @jupi007 said this can be done inside yaru

I also think this idea here is the best solution until gtk4 is available for linux flutter apps as we constantly had problems with handy_window, even though the four rounded corners are very important from a visual point of view

Jupi007:
`handy_window` is also loading a custom stylesheet who mimic Gtk4 look on Gtk3: shadow, bright window border on dark theme, headerbar background color...
If we deprecate `handy_window`, we can move this to `yaru.dart`.

sminez:
Ok, so it sounds like there are a number of things to take into account with if / how we remove `handy_window` in favour of the kind of thing being proposed in this PR. With that in mind, I'd like to revert this PR to the state it was opened in (changes to window creation but `handy_window` still in place) as that at least addresses the immediate issue of the existing approach being completely, consistently broken on some systems.

That change is obviously a hybrid, and not ideal but it at least ensures that the app runs _at all_ on systems where we see this issue. Importantly, the same change needs to be made to the firmware updater as that is unusable on affected systems in its current state.

Once that's sorted we can then try and get to the bottom of what the correct way to do this is in the general case and work towards that for all of the ubuntu desktop flutter applications so we handle things in a consistent way.

@robert-ancell do you see any problems with merging the original state of the PR with this plan in mind?

robert-ancell:
The original behaviour was:

1. Template creates a GtkApplicationWindow
2. handy_window "upgrades" this GtkApplicationWindow to a HdyWindow by modifying the class.

The behaviour after the original PR proposal:

1. Template creates a HdyApplicationWindow
2. handy_window "upgrades" this HdyApplicationWindow to a HdyWindow by modifying the class.

By ensuring the original window is a HdyWindow then this works around any mismatch in the "upgrade" process failing. I hope the "upgrade" in that second step is then essentially a no-op. The other handy_window themeing behaviour would still be applied.

So yes, I think the original PR will probably improve things until we can remove the use of handy_window.

d-loose (approved):

PR #73, opened 2024-09-18
feat: link to new issue templates on 'report issues'

d-loose:
This makes sure we point users that click on the 'report issue' link to the new templates.


spydon (approved):
Haha riiight, because the only `/new` bypasses the choices, I remember that we had that on Flame too and we wondered how people managed to bypass them :joy:

PR #79, opened 2024-10-03
chore: move snap build to core22

sminez:


seb128 (approved):
Technically that one was never on core22 so it's not a revert as claimed by the commit message, but still yes that's what we want now!

Issue #99, opened 2024-12-11
Migrate Desktop Icon to hicolor as a fallback for Future Stepping Stone on building the Desktop Icon on Yaru


ochi12:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is the feature related to a problem or existing issue?

Security Center is now included by default in the latest Ubuntu Desktop release. I believe this is an excellent opportunity to extend support for the Yaru icon theme, at least for the `fullcolor` variant.

### Describe the solution/feature you'd like.

I would like to update the current Security Center desktop icon, which is currently a static reference, to fall back to the `hicolor` theme. This approach would allow me to initiate a new pull request in Yaru to incorporate these new icons.

### Describe any alternatives you've considered.

Approach # 1:
- install the icon to /var/lib/snapd/desktop/icons.

Approach # 2:
- introduce a new usr/share/icons directory just like the implementation on  `g4music` .

ochi12:
@juanruitina

- I moved the svg icon the the root directory
- replaced the name with the convention snap.<app-name>.*
- added this lines to the security-center: part
```yaml
      mkdir -p $CRAFT_PART_INSTALL/meta/gui/icons/hicolor/scalable/apps
      cp snap.desktop-security-center*  $CRAFT_PART_INSTALL/meta/gui/icons/hicolor/scalable/apps
```

build the snap package, installed it and it was successfull.

I have one problem tho...
extra icons are still installed to meta/gui/ .I can't find which part of the source is causing it. as far as I know I already deleted the desktop-security-center
![image](https://github.com/user-attachments/assets/fd6382d7-dcb9-46ad-a61a-0f77ed445835)

PR #100, opened 2024-12-17
migrated desktop icons

ochi12:
Allows desktop icon to be themed

task:
- [ ] solve some icons still installed here

resolves issue #99


ogres:
@ochi12  It has been a while since this PR and lots have changed in Snap. Please reopen if still valid, thanks!

PR #117, opened 2025-03-19
ci: add osv scanner

d-loose:
This adds a workflow that runs [OSV-Scanner](https://google.github.io/osv-scanner/) on pull requests and pushes to `main` using https://github.com/google/osv-scanner-action. The [docs](https://google.github.io/osv-scanner/github-action/) seem to be somewhat outdated, so I adjusted the workflow as needed. On pull requests the `osv-scanner-reusable-pr.yml` workflow from the osv-scanner-action is triggered - it runs the scanner on both the `main` and PR branches and only fails if the PR introduced additional vulnerabilities. On merges to `main` the `osv-scanner-reusable.yml` workflow from the osv-scanner-action is run, which does a full scan of the repository.
As this action covers C/C++, Go, Dart and Rust, I'd propose using it as-is in our other repositories as well.

UDENG-6417

### DIFF
--- a/.github/workflows/osv.yaml
+++ b/.github/workflows/osv.yaml
@@ -1,0 +1,31 @@
+name: OSV-Scanner
+
+on:
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
+
+permissions:
+  actions: read
+  security-events: write
+  contents: read
+
+jobs:
+  scan-push:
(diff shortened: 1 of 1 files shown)


CarlosNihelton (approved):
I can't be sure about using it for Go and Rust right now, but we should definitely give it a try for C, C++ and Dart.

PR #119, opened 2025-04-04
ci(tics): temporarily run on PR

d-loose:


.github/workflows/ci.yaml
```diff
@@ -82,7 +82,7 @@ jobs:
         run: melos build
 
       - name: Run TICS analysis
-        if: github.event_name == 'push'
+        # if: github.event_name == 'push'
```
sergio-costas, line 85: But does in the other cases have access to the TICSAUTHTOKEN?

sergio-costas (commented):
Just a little doubt...

.github/workflows/ci.yaml
d-loose: apparently it does - I just opened this PR to trigger a test run, which seems to have worked

PR #120, opened 2025-04-17
chore: Update to Flutter and dependencies

ashuntu:
Updates Flutter to 3.29.2 and Yaru to 7. Also all upgrades dependencies that satisfy version constraints.

---

UDENG-6719
UDENG-6717


d-loose:
Since the l10n files now need to be checked into the repo, could you add a CI check that ensures they're up to date? (We already have similar ones in the app-center and elsewhere that you can re-use).

d-loose (approved):

PR #126, opened 2025-06-09
feat: replace recovery key flow

matthew-hagemann:
Adds UI for replacing a recovery key

I've aimed to make the providers for this feature as small as possible for easier state management / dependency injection in the unit tests.

`packages/security_center/lib/widgets/file_picker_dialog.dart` and `_RecoveryKeyQRDialog` are lifted from the installer. If its the only other place they live and don't change often, I think its _okay_* to do a little copy if that's okay :sweat_smile: 

## Happy path:
https://github.com/user-attachments/assets/8191b9fc-7938-4ddf-9ca8-6eff8161022e

## Error handling saving a file:
https://github.com/user-attachments/assets/b07c6903-6cd7-4490-9002-3446bb43727c


d-loose (approved):

d-loose (dismissed):

d-loose:
> The title YaruInfo message has excessive line height. I think it's due to [this](https://github.com/ubuntu/yaru.dart/blob/30ea280508f899d5cd8bf66ab3c08c76217ebf7b/lib/src/widgets/yaru_info.dart#L172), 1.8 is way too big, I noticed that already when using YaruInfo in the Firmware Updater. We probably shouldn't set a line height, I'm pretty sure the default/auto one will be good enough. Maybe this was introduced to increase the spacing above and below, but that should be achieved with spacing properties (margin), not by tweaking the line height. Can we work together to fix this in Yaru?

That's already tracked in https://github.com/ubuntu/yaru.dart/issues/995, right?

PR #132, opened 2025-06-23
feat: entropy checks

matthew-hagemann:
Adds call to check pin/passphrase entropy, then display messaging to the user

# Demo

## Passphrase

[Screencast From 2025-06-23 09-41-00.webm](https://github.com/user-attachments/assets/75933669-918b-4c37-b5a1-a94c2286585f)

## Pin

[Screencast From 2025-06-23 09-43-20.webm](https://github.com/user-attachments/assets/a155b714-89e3-4fa6-b230-8e9637f94a75)

### DIFF
--- a/packages/security_center/test/disk_encryption/disk_encryption_page_test.dart
+++ b/packages/security_center/test/disk_encryption/disk_encryption_page_test.dart
@@ -505,9 +505,9 @@ void main() {
     await tester.tap(find.text(tester.l10n.recoveryKeyPinButton));
     await tester.pumpAndSettle();
 
-    // Find all Show buttons (should be 3 - current, new, confirm passphrase)
+    // Find Show button
     final showButtons = find.text(tester.l10n.recoveryKeyPassphraseShow);
-    expect(showButtons, findsNWidgets(3));
+    expect(showButtons, findsOneWidget);
 
     // Find the text fields and enter text
     final textFields = find.byType(TextField);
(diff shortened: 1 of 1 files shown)


packages/security_center/lib/services/disk_encryption_service.dart
```diff
@@ -4,6 +4,19 @@ import 'package:security_center/disk_encryption/disk_encryption_providers.dart';
...
 part 'disk_encryption_service.g.dart';
 
+/// A Class to model an Entropy response from Snapd
+@freezed
+class EntropyResponse with _$EntropyResponse {
+  const factory EntropyResponse({
+    required int entropyBits,
+    required int minEntropyBits,
+    required int optimalEntropyBits,
+  }) = _EntropyResponse;
```
d-loose, line 14: I'd propose using a richer `EntropyResponse` structure, similar to what we're working on in the installer, see https://github.com/canonical/ubuntu-desktop-provision/pull/1085/files#diff-25c90a4ccb338320620a8697e0e979eaa2bed04dab5cc8e1ae4f5774089fc568 from https://github.com/canonical/ubuntu-desktop-provision/pull/1085.

Once the improved API lands, snapd will respond with an error if the check fails, which we should use as a direct indicator instead of performing comparisons on the provided entropy data in the UI.

d-loose (approved):

juanruitina:
@matthew-hagemann The squashing is not the end of the world, but the vertical alignment with the field is off in the first and third input fields.

PR #134, opened 2025-06-26
feat: snapd disk encryption service impl

matthew-hagemann:
## Changes
- Adds `snapd` implementation of `DiskEncryptionService`
- Swap to using `snapd.dart` feature branch
- Swap to using `snapd` types where appropriate in `DiskEncryptionService`
- Add loading state to `replaceRecoveryKey` and `changePinPassphrase`
- `FakeDiskEncryptionService` is now behind the `-a --dry-run` flag
- Drop unused `CheckRecoveryKeyDialogState` from `DiskEncryptionService`
- Drop integration tests
## Demo
[Screencast From 2025-06-26 09-20-36.webm](https://github.com/user-attachments/assets/8d7639d5-0bc5-4cd4-a035-d809ccb46c0d)

### DIFF
--- a/packages/security_center/lib/disk_encryption/disk_encryption_page.dart
+++ b/packages/security_center/lib/disk_encryption/disk_encryption_page.dart
@@ -17,6 +17,8 @@ const _learnMoreUrl =
     'https://discourse.ubuntu.com/t/hardware-backed-encryption-and-recovery-keys-in-ubuntu-desktop/58243';
 const defaultRecoveryKeyFileName = 'recovery-key.txt';
 
+const actionButtonSize = Size(100, 40);
+
 class DiskEncryptionPage extends ConsumerWidget {
   const DiskEncryptionPage({super.key});
 
@@ -236,6 +238,7 @@ class ReplaceRecoveryKeyDialog extends ConsumerWidget {
     );
     final recoveryKey = ref.watch(generatedRecoveryKeyModelProvider);
     final filePicker = ref.read(filePickerProvider);
+    final theme = Theme.of(context);
 
     final l10n = AppLocalizations.of(context);
     return Scaffold(
@@ -370,6 +373,9 @@ class ReplaceRecoveryKeyDialog extends ConsumerWidget {
                   mainAxisAlignment: MainAxisAlignment.end,
                   children: [
                     OutlinedButton(
+                      style: theme.filledButtonTheme.style?.copyWith(
+                        minimumSize: WidgetStateProperty.all(actionButtonSize),
+                      ),
                       onPressed: replaceDialogState
                               is ReplaceRecoveryKeyDialogStateInput
                           ? () => Navigator.of(context).pop()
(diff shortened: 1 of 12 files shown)


packages/security_center/lib/disk_encryption/disk_encryption_page.dart
```diff
@@ -17,6 +17,8 @@ const _learnMoreUrl =
     'https://discourse.ubuntu.com/t/hardware-backed-encryption-and-recovery-keys-in-ubuntu-desktop/58243';
 const defaultRecoveryKeyFileName = 'recovery-key.txt';
 
+const actionButtonSize = Size(100, 40);
```
d-loose, line 20: Is this minimum button size something we're supposed to implement for all buttons? If so, can this be a global theme override? (We should then also implement this in yaru)

packages/security_center/lib/disk_encryption/disk_encryption_providers.dart
```diff
@@ -17,11 +18,7 @@ part 'disk_encryption_providers.g.dart';
 
 final _log = Logger('disk_encryption_providers');
 
-enum SemanticEntropy {
-  belowMin,
-  belowOptimal,
-  optimal;
-}
+enum SemanticEntropy { belowMin, belowOptimal, optimal }
```
d-loose, line 21: Looks like there are some unrelated formatting changes

packages/security_center/lib/main.dart
```diff
@@ -41,31 +39,33 @@ Future<void> main(List<String> args) async {
...
   registerServiceInstance<LocalSnapData>(snapMetadata);
 
   if (argResults.flag('dry-run')) {
     registerService<AppPermissionsService>(
-      () => FakeAppPermissionsService.fromFile(
-        argResults['test-rules'] as String,
-      )..init(),
+      () =>
+          FakeAppPermissionsService.fromFile(argResults['test-rules'] as String)
+            ..init(),
```
d-loose, line 51: Formatting change

packages/security_center/lib/services/fake_disk_encryption_service.dart
```diff
@@ -97,33 +88,30 @@ class FakeDiskEncryptionService implements DiskEncryptionService {
...
   @override
-  Future<void> changePINPassphrase(
+  Future<void> changePinPassphrase(
     AuthMode auth,
     String oldAuth,
     String newAuth,
   ) async {
+    await Future.delayed(
+      const Duration(seconds: 2),
+    ); // Uncomment to simulate a delay
```
d-loose, line 99: The comment suggests you wanted this to be commented out - I don't mind either way, though :)

d-loose (commented):
Thanks, the implementation looks great! Just a few minor nitpicks

packages/security_center/lib/disk_encryption/disk_encryption_page.dart
juanruitina: I'd also be careful with this, yep. Is there any way we can set the size of the button with spinner dynamically, so it preserves the previous size when it had a text label?

juanruitina:
I think no confirmation info panel is needed for changing PIN or passphrase: you can just close the modal once it has updated successfully. We don't with the recovery key because the modal might still be useful so the user can copy it; there's no value here.

Please remember to align with @d-loose on how to align (pun intended) vertically the success/error icons next to the input fields, and the 500ms delay for showing errors, so it's consistent with the installer.

matthew-hagemann:
Thanks @juanruitina!
I've opened a ticket for the 500ms delay, and will open one for automatically dismissing the modal without a confirmation dialog. 

I'll open a PR now for the vertical alignment of the success icons and second loading bar

PR #137, opened 2025-06-30
feat: snapd impl update

matthew-hagemann:
## Changes
- Swap enumerate key slots to expect a map, not a list
- Snapd errors thrown by the entropy check are fatal and handled differently to regular errors
## Screenshot - Snapd errors in entropy checks are fatal
Screenshot From 2025-07-01 14-05-12

### DIFF
--- a/packages/security_center/lib/services/snapd_disk_encryption_service.dart
+++ b/packages/security_center/lib/services/snapd_disk_encryption_service.dart
@@ -63,7 +63,7 @@ class SnapdDiskEncryptionService implements DiskEncryptionService {
   }
 
   @override
-  Future<SnapdEntropyResponse> pinPassphraseEntropyCheck(
+  Future<EntropyResponse> pinPassphraseEntropyCheck(
     AuthMode authmode,
     String newPass,
   ) async {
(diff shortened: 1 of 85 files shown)


packages/security_center/lib/services/snapd_disk_encryption_service.dart
```diff
@@ -67,13 +67,18 @@ class SnapdDiskEncryptionService implements DiskEncryptionService {
...
+        case AuthMode.none:
+          throw UnimplementedError();
+        case AuthMode.pin:
+          return _snapd.checkPin(newPass);
+        case AuthMode.passphrase:
+          return _snapd.checkPassphrase(newPass);
+      }
+    } on SnapdException catch (e) {
+      _log.error(e);
+      rethrow;
```
d-loose, line 81: We shouldn't rethrow the exception here, but handle it and return a richer response that can be handled by the UI.
The issue is that snapd responds with a `400` when the entropy is to low, which is turned into a `SnapdException` by snapd.dart.
I'd propose adding a type that looks like this:
```dart
@freezed
class EntropyResponse with _$EntropyResponse {
  const factory EntropyResponse({
    required bool success,
    required int entropyBits,
    required int minEntropyBits,
    required int optimalEntropyBits,
    List<String>? failureReasons,
  }) = _EntropyResponse;
```

If a `SnapdException` with `kind` `invalid-passphrase` or `invalid-pin` is thrown, we can extract the `entropyBits`, `minEntropyBit`, `optimalEntropyBits` and `failureReasons` from the exception's `value` and return an `EntropyResponse` with `success = false`.

See previous comment [here](https://github.com/canonical/desktop-security-center/pull/132#discussion_r2161289433), which I forgot to follow up on - sorry!

d-loose (changes requested):

matthew-hagemann:
Now correctly catching and transforming snapd errors from entropy checks
[Screencast From 2025-07-02 14-06-59.webm](https://github.com/user-attachments/assets/6112cab7-a800-490e-83bb-c8433aed8a0a)

d-loose (approved):

PR #141, opened 2025-07-17
feat: conditionally hide tpm/fde page

matthew-hagemann:
Interested in your thoughts, but I have added a `featuresService` to act as some source of truth for what features we are enabling. I the use it to build out available routes in the router as well as to decide what services to build and add.
- Added `FeaturesService`
- Moved dry-run ownership into feature service
- `FeatureService` used to determine if we are using encryption by checking the existence of `/dev/disk/by-label/ubuntu-data-enc` and conditionally hiding the page if it doesn't exist (similar check to fwupd: https://github.com/fwupd/fwupd/blob/main/libfwupdplugin/fu-context.c#L1084)

Seems to work as expected, testing on my machine that is encrypted but not using tpm/fde vs my vm that emulates a TPM module

[UDENG-7339](https://warthogs.atlassian.net/browse/UDENG-7339)

[UDENG-7339]: https://warthogs.atlassian.net/browse/UDENG-7339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/security_center/lib/routes.dart
+++ b/packages/security_center/lib/routes.dart
@@ -112,16 +112,24 @@ enum Routes {
     return Routes.fromRoute(route).title(l10n, queryParameters);
   }
 
-  static YaruMasterTileBuilder get tileBuilder => (context, index, selected,
-          availableWidth) =>
-      YaruMasterTile(
-        leading: Icon(availableRoutes[index].icon?.call(selected)),
-        title: Text(availableRoutes[index].title(AppLocalizations.of(context))),
-      );
+  static YaruMasterTileBuilder get tileBuilder =>
+      (context, index, selected, availableWidth) => YaruMasterTile(
+            leading: Icon(availableRoutes[index].icon?.call(selected)),
+            title: Text(
+              availableRoutes[index].title(AppLocalizations.of(context)),
+            ),
+          );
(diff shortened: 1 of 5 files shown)


packages/security_center/lib/routes.dart
```diff
@@ -110,11 +112,30 @@ enum Routes {
...
+    final routes = <Routes>[Routes.appPermissions];
+
+    // Only include disk encryption if the feature is available
+    final featureService = getService<FeatureService>();
+    if (featureService.isDiskEncryptionAvailable) {
+      routes.add(Routes.diskEncryption);
+    }
+
+    return routes;
+  }
```
d-loose, line 140: At this point we should probably create a dedicated `AvailableRoutes` class and instantiate it once with a fixed set of routes. As a static method this currently checks whether disk encryption is available every time there is a widget rebuild. Doing this only once when the app starts should be enough.

[Screencast From 2025-07-18 14-39-23.webm](https://github.com/user-attachments/assets/e07aaca1-6d75-409b-9fcb-d19d2ecea9a9)

d-loose (approved):

PR #149, opened 2025-07-29
feat: fde status banners & error handling

matthew-hagemann:
- Adds status banners for TPM enabled and what AuthMode is in use (if any)
- Adds more explicit error banners based on what went wrong
- Fixes a bug in identifying `authMode` where we were looking at the wrong keyslot, which was missed until now due to fallback values I've now removed
- Unifies everything under one progress indicator:
  - To accomplish this, I did refactor a fair amount of the `disk_encryption_page.dart` widgets to hold everything that under one body that reacts the same `TpmAuthenticationModel` providers for error states, status banner etc.

### PIN in use
Screenshot From 2025-07-30 08-57-20

### Passphrase in use
Screenshot From 2025-07-30 08-57-54

### No authmode in use
Screenshot From 2025-07-30 08-59-51

### General error when calling to `EnumerateKeyslots`
Screenshot From 2025-07-30 09-13-53

### `snapd` 404's on `EnumerateKeyslots` => unsupported `snapd` version
Screenshot From 2025-07-30 08-33-15

### `snapd` 401 errors => don't have access to the plug we need
image

## Unified progress indicator
[Screencast From 2025-07-30 09-25-24.webm](https://github.com/user-attachments/assets/09ce3279-6b0f-40e9-abaa-f5bca917c803)


packages/security_center/lib/disk_encryption/disk_encryption_providers.dart
```diff
@@ -283,10 +326,17 @@ class TpmAuthenticationModel extends _$TpmAuthenticationModel {
...
-    } on Exception catch (e) {
+
+      return AuthMode.none;
+    } on SnapdException catch (e) {
       _log.error('Failed to determine TPM authentication mode: $e');
-      return AuthMode.passphrase;
+      if (e.statusCode == 404) {
+        throw SnapdStateExceptionUnsupportedSnapdVersion();
+      }
+      if (e.statusCode == 401) {
```
matthew-hagemann, line 336: Disconnecting `snap-fde-control` didn't result in any issue for me, so I am not certain this is the correct error code.

packages/security_center/lib/disk_encryption/disk_encryption_providers.dart
```diff
@@ -283,10 +326,17 @@ class TpmAuthenticationModel extends _$TpmAuthenticationModel {
...
-    } on Exception catch (e) {
+
+      return AuthMode.none;
+    } on SnapdException catch (e) {
       _log.error('Failed to determine TPM authentication mode: $e');
-      return AuthMode.passphrase;
+      if (e.statusCode == 404) {
+        throw SnapdStateExceptionUnsupportedSnapdVersion();
+      }
+      if (e.statusCode == 403) {
```
matthew-hagemann, line 336: From testing, 403 seems to be the correct error code when the interface is not connected?

packages/security_center/lib/disk_encryption/disk_encryption_page.dart
```diff
@@ -33,109 +33,165 @@ class _Body extends ConsumerWidget {
...
+          ),
+          const SizedBox(height: 16),
+          Text(
+            l10n.diskEncryptionPageRecoveryKey,
+            style: Theme.of(context).textTheme.titleMedium,
+            textAlign: TextAlign.left,
+          ),
+          const SizedBox(height: 16),
+          MarkdownText(
+            '${l10n.diskEncryptionPageStoreYourKey} ${l10n.diskEncryptionPageLearnMore.link(_learnMoreUrl)}',
```
d-loose, line 103: I think concatenating l10n strings like this will potentially case issues with RTL languages.

packages/security_center/lib/disk_encryption/disk_encryption_page.dart
```diff
@@ -33,109 +33,165 @@ class _Body extends ConsumerWidget {
...
+                    child: Text(l10n.recoveryKeyPassphraseButton),
+                  ),
+                ],
+              ),
+            _ => const SizedBox.shrink(),
+          },
+        ],
       ),
+      error: (e, stack) {
+        try {
```
d-loose, line 173: Can this just be a `switch` expression? I don't think throwing and immediately catching the exception is necessary here

juanruitina:
Two minor comments:

- [ ] The heading with the passphrase options should be "Encryption passphrase"
- [ ] Please use the monospace font for the command in the "Security Center cannot connect to the snapd interface" error message

When we are loading the whole page, we should probably be using the spinner, but we can decide on that when we do the full UX review of the feature. Other than that, this all looks good to me.

packages/security_center/test/test_utils.dart
```diff
@@ -133,31 +133,81 @@ DiskEncryptionService registerMockDiskEncryptionService({
...
   AuthMode authMode = AuthMode.pin,
   EntropyResponse Function(String)? entropyResponseBuilder,
+  bool enumerateKeySlots404Error = false,
+  bool enumerateKeySlots403Error = false,
+  bool enumerateKeySlotsFailure = false,
+  bool missingRecoveryKeySlot = false,
+  bool missingDefaultKeySlot = false,
+  bool missingDefaultFallbackKeySlot = false,
+  bool invalidTpmPlatformName = false,
+  bool authModeMismatch = false,
```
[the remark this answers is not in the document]
matthew-hagemann: yea, this definitely feels wrong :joy: thanks!

PR #153, opened 2025-08-27
fix: follow rename of Settings › Applications to Apps

jbicha:
The gnome-control-center panel is named `Apps` (although it previously was named `Applications`).

I updated the English translations but I didn't touch the other translations.


ogres:
Thanks for the contribution, looks like this was already changed on main branch. I will close the PR - thanks!

PR #157, opened 2025-09-18
feat: use local icons from snapd instead of fetching them from the network

d-loose:
This makes use of https://github.com/canonical/snapd.dart/pull/130 to retrieve icons for locally installed snaps without the need for networks requests.
I've also removed the `cached_network_image` dependency and the custom cache manager that's no longer needed.

UDENG-7930


matthew-hagemann (approved):
Thanks for this!

PR #159, opened 2025-09-26
feat(app_permissions): adding camera interface support

matthew-hagemann:
Adds support for the camera interface behind the ` -a --enable-camera-interface` or `-a --dry-run` feature flags while we wait for support to land in `snapd`

Integration tests are tricky as we rely on calling for snaps on the system with the camera interface. Will focus on these in the e2e tests in my next ticket.

# Update post UI/UX review
We may update this based on UX explorations to use a drop down to better expose all possible lifespans and rules assigned to each snap.

We will also create a follow up PR to standardize `ListTile` heights and padding across the security center.

# Interfaces page
## Snaps with camera interface present
Screenshot From 2025-09-30 13-22-11

## No snaps present
Screenshot From 2025-09-30 13-21-05

# Snaps page
## Snaps with camera interface present
Screenshot From 2025-09-30 13-22-16
Screenshot From 2025-09-30 13-22-23

## No snaps present
Screenshot From 2025-09-30 13-21-11


packages/security_center/lib/app_permissions/snaps_page.dart
```diff
@@ -90,3 +163,45 @@ class _AppTile extends ConsumerWidget {
...
+      child: ListTile(
+        leading: AppIcon(
+          snapIcon: ref.watch(snapIconProvider(snapName)),
+        ),
+        title: Text(ref.watch(snapTitleOrNameProvider(snapName))),
+        trailing: Switch(
+          value: hasAllowRule,
+          onChanged: (value) {
+            if (value) {
+              notifier.removeAll().then((_) => notifier.createAccessRule());
```
matthew-hagemann, line 198: Just incase some odd state is created external to the security center, I have it clear all rules before we add an allow all rule.

packages/security_center/lib/app_permissions/rules_providers.dart
```diff
@@ -31,39 +37,72 @@ Future<List<SnapdRule>> rules(Ref ref) async =>
       _ => [],
     };
 
+@riverpod
+Future<List<String>> cameraSnaps(Ref ref) async {
+  final service = getService<AppPermissionsService>();
+  return service.getSnapsWithInterface('camera');
```
matthew-hagemann, line 43: Added this to the snapd service to retrieve any installed snaps with the camera interface, which we use to then track which snaps to allow the user to write / remove access rules for

packages/security_center/lib/app_permissions/app_rules_page.dart
```diff
@@ -31,6 +31,7 @@ class AppRulesPage extends ConsumerWidget {
             error: (error, _) => ErrorWidget(error),
             loading: () => const Center(child: YaruCircularProgressIndicator()),
           ),
+      SnapdInterface.camera => const SizedBox.shrink(),
```
matthew-hagemann, line 34: As we are not looking for the same level of rules granularity, I've not added any code to the `app_rules_page` for the camera interface, and handle everything instead in the `snaps_page`, which now has logic to build differently depending on if we are working with the camera of home interfaces.

packages/security_center/lib/app_permissions/snaps_page.dart
```diff
@@ -24,7 +24,14 @@ class SnapsPage extends ConsumerWidget {
       data: (snapRuleCounts) =>
           _Body(snapRuleCounts: snapRuleCounts, interface: interface),
       error: (error, _) => ErrorWidget(error),
-      loading: () => const Center(child: YaruCircularProgressIndicator()),
+      loading: () {
+        // Keep showing previous data if available to prevent flickering
+        final previousData = snapRuleCounts.valueOrNull;
+        if (previousData != null) {
+          return _Body(snapRuleCounts: previousData, interface: interface);
+        }
```
d-loose, line 32: I think the same can be achieved by setting `skipLoadingOnReload` to `true`

matthew-hagemann: Ah, thanks!

packages/security_center/lib/app_permissions/snaps_page.dart
matthew-hagemann: @d-loose I can't see it happening via any of our apps, but my thinking is that technically there are other types of rules that could be written externally by the user we are not interested in for what we are showing in this abstracted view.

My thinking here is, just in-case something external was done by the user that was not `allow:always` we clear it to enforce our abstraction.

So the "odd state" would be, what do we do if externally the user has written a `deny` of any type or a `allow` that is not scoped to always.

We could do an error state where we check to see if there are rules we didn't create and thrown an exception when building the page?

matthew-hagemann: Ah yes, I had missed that, and that also explains the "Reset all permissions" button I wanted to ask Juan about! Thanks, let me update this to match that :slightly_smiling_face:

matthew-hagemann:
#Updates from UI review:
## Interfaces page:
### Moved camera interface under home
Screenshot From 2025-10-03 14-46-53
## Snaps page:
### Added reset button, increased Tile height to allow for the subtitle if a "session" lifespan:
Screenshot From 2025-10-03 14-47-16

PR #169, opened 2025-11-21
feat: pin / passphrase management via replaceProtectedKeyslot snapd api

matthew-hagemann:
Makes use of the the newly supported [replacePlatformKey](https://github.com/canonical/snapd.dart/pull/134) api on snapd to allow the user to swap between pins, passphrases and no authentication by replacing the platform key in use.

The async snapd methods for adding and removing PIN/Passphrases have a bit of a delay. To avoid the user exiting the dialog and us not having an easy way to invalidate page state to force a rebuild when the auth mode is changed, we now have a new loading flow in use for the adding / removing flows:

[Screencast From 2025-11-28 09-50-07.webm](https://github.com/user-attachments/assets/3986708b-e518-4ec3-be47-1198060e4da0)

### DIFF
--- a/packages/security_center/lib/disk_encryption/disk_encryption_page.dart
+++ b/packages/security_center/lib/disk_encryption/disk_encryption_page.dart
@@ -692,9 +692,7 @@ class _NoneAuthenticationActions extends ConsumerWidget {
       children: [
         Text(
           l10n.diskEncryptionPageAdditionalSecurityHeader,
-          style: Theme.of(context).textTheme.titleMedium?.copyWith(
-                fontWeight: FontWeight.bold,
-              ),
+          style: Theme.of(context).textTheme.titleSmall,
         ),
         const SizedBox(height: 8),
         Text(
@@ -763,9 +761,7 @@ class _PassphraseAuthenticationActions extends ConsumerWidget {
       children: [
         Text(
           l10n.recoveryKeyEncrpytionPassphraseHeader,
-          style: Theme.of(context).textTheme.titleMedium?.copyWith(
-                fontWeight: FontWeight.bold,
-              ),
+          style: Theme.of(context).textTheme.titleSmall,
         ),
         const SizedBox(height: 8),
         Text(l10n.recoveryKeyPassphraseBody),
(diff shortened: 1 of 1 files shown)


packages/security_center/lib/disk_encryption/disk_encryption_page.dart
```diff
@@ -47,126 +47,378 @@ class EncryptionPageBody extends ConsumerWidget {
...
+    );
+    ref.listen<RemoveAuthState>(
+      removeAuthModelProvider(AuthMode.passphrase),
+      (previous, next) {
+        if (previous is RemoveAuthStateLoading && next is RemoveAuthStateIdle) {
+          ref.invalidate(tpmAuthenticationModelProvider);
+        }
+      },
+    );
+
```
d-loose: Can't we just add `ref.invalidate(tpmAuthenticationModelProvider)` to `removeAuth()` in the model?

d-loose:
When running this in dry-run mode, there's an additional delay (without a loading indicator) due to rebuilding the `tpmAuthenticationModel` after removing the PIN:

[Screencast From 2025-11-25 10-31-03.webm](https://github.com/user-attachments/assets/6d7fdbb2-881d-44b3-9f53-c05fd81e0ee7)

This is because invalidating the provider changes its state to `AsyncData<AuthMode>(isLoading: true, value: AuthMode.pin)` instead of `AsyncLoading<AuthMode>()`

juanruitina (changes requested):
Just updated some strings. Remaining bits:

- [ ] There should be a single "Learn more" link, under the TPM explanation below the status (I already updated the URL itself). Please add that link and remove all others.
- [ ] When adding PIN/passphrase, the recovery key line should be in a separate paragraph.

I tweaked the designs to match this.

d-loose (approved):
Great work, I think this is a huge improvement!

juanruitina:
We discussed fixing:

- Removing the link under "Additional security"
- Making links focusable

I think we can merge once these two are addressed.

A few more things we found when reviewing, related to styles that should likely be addressed elsewhere or even upstream in Yaru.dart:

- The checkbox icon we show next to input fields with valid inputs is sometimes misaligned. We may want to take a different approach.
- Error message in input fields should be same size as descriptive text (is larger at the moment)
- Same with tile font size, should be labelLarge with descriptive text be labelMedium.

We can tackle these separately.

PR #173, opened 2026-01-06
fix: multiple leftover ui tweaks

matthew-hagemann:
This PR rounds up some of the outstanding UI tweaks from last year:

## The checkbox icon we show next to input fields with valid inputs is sometimes misaligned.
Screenshot From 2026-01-06 14-50-17
Screenshot From 2026-01-06 14-50-35

## Error message in input fields should be same size as descriptive text
Screenshot From 2026-01-06 14-51-51

## Tile font size changes: use `labelLarge` for main text and `labelMedium` for subtitles. Icons need to be shrunk on the `app_permissions` page
Screenshot From 2026-01-06 14-52-20
Screenshot From 2026-01-06 14-52-24

## Links are focusable when tab cycling
[Screencast From 2026-01-06 14-55-21.webm](https://github.com/user-attachments/assets/501a4ccf-02c6-494f-ada6-d2219d983c39)

### DIFF
--- a/packages/security_center/integration_test/security_center_test.dart
+++ b/packages/security_center/integration_test/security_center_test.dart
@@ -1,7 +1,6 @@
 import 'dart:convert';
 import 'dart:io';
 
-import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
 import 'package:security_center/app_permissions/home_interface.dart';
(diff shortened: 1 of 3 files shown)


d-loose (approved):
Just a small nitpick, feel free to ignore for now :)

juanruitina:
Thank you for working on this.

We should use [focus rings](https://github.com/ubuntu/yaru.dart/issues/979) for the focus state. Changing the background is not good for accessibility. Maybe check with @ashuntu, he's been working on focus states on yaru.dart.

I'm curious about the error messages and description in input fields. Do they have any letter spacing set? If so, we should remove it; I did it [upstream](https://github.com/ubuntu/yaru.dart/pull/1036) but I don't know if it applied everywhere.

About tiles: title should have a regular font weight (it looks medium at the moment), and description should be labelMedium (it seems to have the same size in the tiles used in App Permissions).

About headings: can you make the one in App Permissions bodyMedium, regular font weight as you did for Disk Encryption?

matthew-hagemann:
> Thanks, LGTM! I assume the short/mid-term plan is to make the upstream `YaruListTile` match `SecurityCenterListTile`?

Yes :slightly_smiling_face:

PR #179, opened 2026-01-28
fix: drop state mutations for authMode in changeAuthDialog's

matthew-hagemann:
I dropped the setters from the `changeAuthDialog` and moved to using different provider families for each `authMode`. Run this through in a vm and all seems working as intended now.


d-loose (approved):
Thanks, this looks good!
Looking at the existing tests, it might be a good idea to do some assertions on the auth mode in the mocks. Currently the mocked `pinPassphraseEntropyCheck` and `replacePlatformKey` provide valid responses even if wrong arguments are passed to them, which might be the reason we didn't catch this earlier

PR #187, opened 2026-02-16
feat: add check for snapd microphone support

gabelluardo:
This gates the microphone feature behind a check for `snapd` support for `microphone-interface`

I've added `pub_semver` for semantic version comparison, maintained by the [dart team](https://github.com/dart-lang/tools/tree/main/pkgs/pub_semver), so it should be fine

---
[UDENG-9107](https://warthogs.atlassian.net/browse/UDENG-9107)

[UDENG-9107]: https://warthogs.atlassian.net/browse/UDENG-9107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/security_center/test/app_permissions/interfaces_page_test.dart
+++ b/packages/security_center/test/app_permissions/interfaces_page_test.dart
@@ -112,7 +112,7 @@ void main() {
         );
         expect(homeInterfaceTile, findsOneWidget);
 
-        // Microphone interface should be hidden
+        // Microphone interface visibility should match expected behavior
         final microphoneInterfaceTile = find.ancestor(
           of: find.text(SnapdInterface.microphone.localizedTitle(tester.l10n)),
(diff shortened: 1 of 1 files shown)


packages/security_center/lib/services/feature_service.dart
```diff
@@ -37,6 +41,10 @@ class FeatureService {
     return _isUsingFde!;
   }
 
+  bool get supportsMicrophone {
+    return _hasGreaterVersion(snapdVersion, '2.75');
```
matthew-hagemann: I think this should first check if we are in `dry-run` mode. If we are, its okay to show, as we are using the mock back end, and allows us to test. So similar to what is being done in `isDiskEncryptionAvailable` just above

PR #197, opened 2026-03-24
fix: Add 'pro' to .desktop keywords

ashuntu:
Title.


ashuntu:
I assume we do not need to add "Ubuntu Pro" as a separate keyword as "Ubuntu" and "Pro" should cover that?

d-loose (approved):

PR #201, opened 2026-04-10
feat: Show an error if Ubuntu Pro detach fails

ashuntu:
Detach errors are very unlikely and we can't obtain much information about the cause of failure, but previously they were silently dropped. This shows a generic error message in the detach dialog if detaching fails.

image

---

UDENG-9316

### DIFF
--- a/packages/security_center/lib/l10n/app_en.arb
+++ b/packages/security_center/lib/l10n/app_en.arb
@@ -290,6 +290,7 @@
     "ubuntuProDisablePro": "Disable Ubuntu Pro",
     "ubuntuProDisable": "Disable",
     "ubuntuProDisablePrompt": "Disabling Ubuntu Pro will detach your subscription from this machine. Do you want to proceed?",
+    "ubuntuProDisableError": "Could not disable Ubuntu Pro. Please try again.",
     "ubuntuProEnable": "Enable",
     "ubuntuProCancel": "Cancel",
     "ubuntuProFeatureEnableError": "Could not enable the feature, please try again.",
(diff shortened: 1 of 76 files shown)


ashuntu:
@juanruitina Design input on this? Not sure if we have a similar flow elsewhere that we need to display errors for. We don't really have a way to determine an actual error, so the message is somewhat generic.

juanruitina:
@ashuntu Even if we can't communicate it in every specific instance, do we know the reasons why this happens? It can help me write a more useful error message, even if not totally tailored.

ashuntu:
> @ashuntu Even if we can't communicate it in every specific instance, do we know the reasons why this happens? It can help me write a more useful error message, even if not totally tailored.

I've never actually been able to get detaching to fail (other than on purpose by failing the polkit prompt), but in theory it can. I would assume the most common would be invalid auth or the dbus service is busy/times out in some way. Software Properties also displays a generic error string almost identical to the one I've added here.

juanruitina:
@ashuntu Then just use this (no period)

> Could not disable Ubuntu Pro, try again

In fact, it would be great if you remove every "please" from every string as to align with [content guidelines](https://vanillaframework.io/docs/content-guidelines).

d-loose (dismissed):
Thanks, LGTM!
Can you coordinate with @matthew-hagemann concerning the merge conflict? He'll have a look at the weblate PR that also needs rebasing. Afterwards we can re-enabled translations on weblate

matthew-hagemann:
> Thanks, LGTM! Can you coordinate with @matthew-hagemann concerning the merge conflict? He'll have a look at the weblate PR that also needs rebasing. Afterwards we can re-enabled translations on weblate

Working on this now :+1:

ashuntu:
Rebased on main and resolved the translation conflicts.

d-loose (approved):

PR #220, opened 2026-05-15
Fix KDE Plasma taskbar icon matching

Iamliuxiaozhen:
## Summary

KDE Plasma may show a generic Wayland placeholder icon for Security Center in the taskbar instead of the correct application icon.

This happens because the running window reports:

* `desktopFile: security_center`
* `resourceClass: security_center`
* `resourceName: security_center`

but the desktop entry does not declare a matching `StartupWMClass`.

This PR adds:

```ini
StartupWMClass=security_center
```

so Plasma can correctly associate the running window with its desktop entry and icon.

## Testing

Tested locally on:

* KDE Plasma
* Wayland
* Ubuntu Studio 26.04

Verified that:

* the correct taskbar icon is shown
* the generic Wayland placeholder icon no longer appears
* launcher/taskbar grouping works correctly

### DIFF
--- a/snap/gui/desktop-security-center.desktop
+++ b/snap/gui/desktop-security-center.desktop
@@ -2,7 +2,7 @@
 Type=Application
 Version=1.0
 Exec=desktop-security-center %U
-Icon=${SNAP}/meta/gui/security-center.svg
+Icon=${SNAP}/meta/gui/desktop-security-center.png
 StartupWMClass=security_center
 Terminal=false
 Categories=System;Utility;Permissions;Settings;Security;
--- a/snap/gui/desktop-security-center.desktop
+++ b/snap/gui/desktop-security-center.desktop
@@ -2,7 +2,8 @@
 Type=Application
 Version=1.0
 Exec=desktop-security-center %U
-Icon=${SNAP}/meta/gui/desktop-security-center.png
+Icon=${SNAP}/meta/gui/security-center.svg
+StartupWMClass=security_center
 Terminal=false
 Categories=System;Utility;Permissions;Settings;Security;
 Keywords=Ubuntu;Permissions;Snaps;Security;Pro;


Iamliuxiaozhen:
Friendly ping 🙂

I tested this on KDE Plasma Wayland and it fixes the taskbar icon matching issue caused by the missing StartupWMClass entry.

Happy to adjust the implementation if needed. Thanks!

snap/gui/desktop-security-center.desktop
```diff
@@ -2,7 +2,8 @@
 Type=Application
 Version=1.0
 Exec=desktop-security-center %U
-Icon=${SNAP}/meta/gui/desktop-security-center.png
+Icon=${SNAP}/meta/gui/security-center.svg
```
sergio-costas: Since this PR is replacing the png with the svg, maybe we should also remove the, now unused, png. What do you say, @matthew-hagemann ?

sergio-costas (changes requested):

snap/gui/desktop-security-center.desktop
matthew-hagemann: Hey, testing this on Gnome seems to lead to a regression where no icon now renders for me.
image
image

sergio-costas: Yes, there is a typo...

```suggestion
Icon=${SNAP}/meta/gui/desktop-security-center.svg
```

sergio-costas (dismissed):

matthew-hagemann (dismissed):
Thank you for the PR! 

Image

Surfacing this as a more than a comment: Testing on Gnome, I see this leads to a regression where I am no longer getting an icon for the security center.

snap/gui/desktop-security-center.desktop
sergio-costas: (also, confirmed that the wmclass is `desktop_center`)

Iamliuxiaozhen:
Thanks for testing.

My main issue is that KDE Plasma shows a generic placeholder icon in the taskbar because the running window WM_CLASS does not match the desktop entry.

The Icon field change was not necessary and caused a GNOME regression, so I reverted it.

The fix should only add:

StartupWMClass=security_center

I confirmed this matches the WM_CLASS reported by the running window on KDE Plasma Wayland.

Iamliuxiaozhen:
I verified the window metadata on KDE Plasma Wayland:

resourceClass: security_center
resourceName: security_center
desktopFile: security_center

So StartupWMClass=security_center is correct.

I also checked the snap contents and found that the snap provides:

* desktop-security-center.png
* desktop-security-center.svg

but not security-center.svg.

I've reverted the icon change and kept only the StartupWMClass addition.

Iamliuxiaozhen:
I verified that the running window reports resourceClass: security_center and resourceName: security_center on KDE Plasma Wayland.

However, during local testing I noticed that the icon is shown correctly during startup, then changes back to the generic placeholder once the main window appears. This looks like the startup notification initially matches the desktop entry, but the final Wayland window is still not being associated correctly.

ogres (approved):

PR #221, opened 2026-05-19
docs: AGENTS.md bootstrap

matthew-hagemann:
I want to bootstrap some layered AGENTS.md files for when working with OpenCode / Copilot

Claude only respects CLAUDE.md, and the best solution just seems to be a symlink for now.


AGENTS.md
```diff
@@ -0,0 +1,41 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
```
gabelluardo: I would use a generic `code agent` here and put `Claude` as an e.g. to be model agnostic and avoiding agent identity crisis / hallucination

packages/security_center/test/AGENTS.md
```diff
@@ -0,0 +1,47 @@
+# CLAUDE.md — Tests
```
gabelluardo: Does Claude really need to specify `CLAUDE.md` in each file? Can't this also be generalized to `AGENT.md` or something?

packages/security_center/lib/AGENTS.md
```diff
@@ -0,0 +1,47 @@
+# CLAUDE.md — Feature pages & providers
```
gabelluardo: ditto

packages/security_center/lib/services/AGENTS.md
```diff
@@ -0,0 +1,42 @@
+# CLAUDE.md — Services
```
gabelluardo: ditto

gabelluardo (changes requested):
LGTM, just a couple of comments about agent naming :)

gabelluardo (approved):
Thanks!

PR #223, opened 2026-05-19
feat: retry indeterminate encryption status

matthew-hagemann:
Adds a 2 min exponential retry to the encryption status API so we can see if it resolves beyond indeterminate. See also:
https://bugs.launchpad.net/snapd/+bug/2147606

I iterated on the AGENTS.md's as I ran into issues, which is why they've been updated in parallel

Tested e2e on Resolute

# Demos

## Happy path:
https://github.com/user-attachments/assets/512c95f2-37ab-48f9-b843-7657e8a22837

## Error state is cached with a keep alive so we don't retry if we navigate back:
https://github.com/user-attachments/assets/8c0d2806-25e8-455f-9b6c-702182367573

## We don't resolve to a new status after the timer expires:
https://github.com/user-attachments/assets/4b272ea9-fc6a-4c4b-ae40-c678da31961e

UDENG-9854


packages/security_center/test/disk_encryption/disk_encryption_page_test.dart
```diff
@@ -1094,6 +1107,14 @@ void main() {
 
     for (final tc in cases) {
       testWidgets(tc.name, (tester) async {
+        final prevMaxRetry = TpmAuthenticationModel.maxRetryDuration;
+        final prevInitialDelay = TpmAuthenticationModel.initialRetryDelay;
+        TpmAuthenticationModel.maxRetryDuration = Duration.zero;
+        TpmAuthenticationModel.initialRetryDelay = Duration.zero;
+        addTearDown(() {
```
matthew-hagemann, line 1114: Need to reset them because they are statics

packages/security_center/test/disk_encryption/disk_encryption_page_test.dart
```diff
@@ -1709,4 +1734,54 @@ void main() {
...
+      ),
+    ];
+
+    for (final tc in cases) {
+      test(tc.name, () async {
+        final prevMaxRetry = TpmAuthenticationModel.maxRetryDuration;
+        final prevInitialDelay = TpmAuthenticationModel.initialRetryDelay;
+        TpmAuthenticationModel.maxRetryDuration = tc.maxRetryDuration;
+        TpmAuthenticationModel.initialRetryDelay = Duration.zero;
+        addTearDown(() {
```
matthew-hagemann, line 1762: Same, reset them as they are statics

d-loose (approved):

PR #224, opened 2026-05-21
Allow granular control for permission for each app

SoumyaRanjanPatnaik:
## Description
Replace the camera/microphone allow/deny switch with a popup menu that reflects the actual snapd rule state and lets users change it without losing the current rule semantics.

Fixes UDENG-9874

## What changed
- Added `PermissionRuleMenuAction` and `PermissionRulePopupMenu` under `lib/app_permissions/`.
- Mapped snapd `outcome` + `lifespan` values to user-facing menu states.
- Updated camera and microphone tiles to show the current rule state in the menu button.
- Replaced the old binary switch flow with a menu-driven update flow.
- Kept `Ask always` as the reset/default state, removing rules when selected.
- Added localized labels for temporary allow/deny states.

## Updated UI screenshot
image


packages/security_center/lib/app_permissions/rules_popup_menu.dart
```diff
@@ -0,0 +1,136 @@
...
+enum PermissionRuleMenuAction {
+  foreverAllowed,
+  sessionAllowed,
+  temporarilyAllowed,
+  foreverDenied,
+  sessionDenied,
+  temporarilyDenied,
+  askAlways;
+
+  /// Creates an approriprate instance [PermissionRuleMenuAction] from a snapd outcome and lifespan.
```
matthew-hagemann: approriprate -> appropriate

packages/security_center/lib/app_permissions/rules_popup_menu.dart
```diff
@@ -0,0 +1,136 @@
+import 'package:flutter/material.dart';
+import 'package:security_center/l10n.dart';
+import 'package:snapd/snapd.dart';
+import 'package:yaru/yaru.dart';
+
+/// Defines the possible actions in the rules popup menu, which allows updating
+/// the permission outcome and lifespan for an app's rules.
+enum PermissionRuleMenuAction {
```
matthew-hagemann: This enum is very close to the one we have in `lib/app_permissions/rules_category.dart`. I think this would better if this lives in that file by extending its enum to handle all your use cases, and moving these functions for converting types and to l10n strings there. 

Might also be a good idea to rename that file / its contents as we make it broader. 

I'm thinking: 
- `lib/app_permissions/rules_category.dart` -> `lib/app_permissions/rules_action.dart`
- `SnapdRuleCategory` -> `SnapdRuleAction`

matthew-hagemann (changes requested):
Awesome work! One small thing re: code re-use, but otherwise LGTM!

juanruitina:
Some feedback:

- [x] I don't think we should hide the selected option from the dropdown, even if it's slightly redundant.
- [x] It seems that when you hit enter on the row, screen reader says nothing about the open popup; I think focus should move to the first option. Please check keyboard navigation and screen reading, perhaps with @ashuntu or @paul-geoghegan.
- [x] One minor visual quirk: the button label seems to have the medium font weight, instead of regular (which would be thinner).

packages/security_center/lib/app_permissions/rules_popup_menu.dart
SoumyaRanjanPatnaik: Can you verify the changes now?

paul-geoghegan:
@juanruitina is correct. It should focus the first item when opened.

matthew-hagemann (dismissed):
Awesome work! LGTM 🙂

SoumyaRanjanPatnaik:
@juanruitina @paul-geoghegan I've updated the code to have the first item in focus on menu open. Tested with `orca` and seems to work well. 

Approximate flow:
1. Tabbing into a app tile reads as "Firefox. Allow Always. Update Permissions Button".
2. Space or enter opens the menu and reads "Popup Menu Panel. Ask Always".

juanruitina:
Great improvement, just some comments on the dropdown menu states:

image

- [x] Focus (in the screenshot on the first option) should be a [focus ring](https://github.com/ubuntu/yaru.dart/issues/979) instead of the current background change. @ashuntu did we ever fix this in the Yaru component?
- [x] Ideally, focus ring on autofocus should only be visible if user is indeed using the keyboard, i.e. when user moves to the menu by hitting enter, not when clicking on the button
- [x] Selected (last option) should better be indicated with a checkbox instead of the current background change, GNOME style: 

image

All this would apply to the Yaru components, more than this specific implementation.

ashuntu:
> Focus (in the screenshot on the first option) should be a https://github.com/ubuntu/yaru.dart/issues/979 instead of the current background change. @ashuntu did we ever fix this in the Yaru component?

I don't think so. @SoumyaRanjanPatnaik you might be able to try wrapping the items with `YaruFocusBorder`, but this should be solved in yaru.dart long-term.

SoumyaRanjanPatnaik:
Somehow managed to address the focus ring related comments (with the help of AI ofcourse). The focus ring now appears on keyboard interaction only and disappears when interacted via mouse. I've seperated out the component for the `YaruPopupMenuItem` for now. @matthew-hagemann @ashuntu do you think this should go into the `yaru` package?

matthew-hagemann:
> Somehow managed to address the focus ring related comments (with the help of AI ofcourse). The focus ring now appears on keyboard interaction only and disappears when interacted via mouse. I've seperated out the component for the `PopupMenuItem` for now. @matthew-hagemann @ashuntu do you think this should go into the `yaru` package?

@SoumyaRanjanPatnaik Awesome, thanks! Yes to upstreaming, but I'd land the work here first, then propose the change upstream with this as motivation as why we want to upstream it. 

Will review impl shortly :slightly_smiling_face:

packages/security_center/lib/widgets/yaru_popup_menu_item.dart
```diff
@@ -0,0 +1,199 @@
...
+        contentPadding: EdgeInsets.zero,
+        titleTextStyle: style,
+        child: item,
+      ),
+    );
+
+    return ValueListenableBuilder(
+      valueListenable: _showFocusBorder,
+      builder: (context, showFocusBorder, child) {
+        return AnimatedContainer(
```
matthew-hagemann: Could we not just re-use the `yaru_focus_border` here? This impl looks [close to the same](https://github.com/ubuntu/yaru.dart/blob/main/lib/src/widgets/yaru_focus_border.dart)?

packages/security_center/lib/widgets/yaru_popup_menu_item.dart
```diff
@@ -0,0 +1,199 @@
...
+    final theme = Theme.of(context);
+    final popupMenuTheme = PopupMenuTheme.of(context);
+    final states = <WidgetState>{
+      if (!widget.enabled) WidgetState.disabled,
+    };
+    final style = _textStyle(context, states);
+    final padding = widget.padding ??
+        (theme.useMaterial3
+            ? const EdgeInsets.symmetric(horizontal: 12)
+            : const EdgeInsets.symmetric(horizontal: 16));
```
matthew-hagemann: Yaru will always be `theme.useMaterial3` no? I don't think we need check this.

packages/security_center/lib/widgets/yaru_popup_menu_item.dart
```diff
@@ -0,0 +1,199 @@
...
+        ),
+      ],
+    );
+  }
+
+  TextStyle _textStyle(BuildContext context, Set<WidgetState> states) {
+    final theme = Theme.of(context);
+    final popupMenuTheme = PopupMenuTheme.of(context);
+
+    if (theme.useMaterial3) {
```
matthew-hagemann: Yaru will always be `theme.useMaterial3` no?

matthew-hagemann (changes requested):
Couple small nits from the new widget. Sorry for the delay, took me some time to gather my thoughts based on the `yaru.dart` codebase.

The custom handling around the input handling for the widget is not really something we normally do. I can't say its wrong, but it does seem like we are fighting Flutter/Yaru a bit, but I can't see how else you achieve the focus ring appearing on keyboard interaction only and disappears when interacted via mouse. 

It would mean this widget handles differently from others in Yaru, but that could be more we have work to do in Yaru than this is wrong. @ashuntu not sure if you have any thoughts?

SoumyaRanjanPatnaik:
Very valid comments haha. I was trying to make this work myself and found myself fighting with Yaru. Asked opencode to fixed it and it spit out this code. Will make the changes required. Don't know how I missed the material3 stuff. 

Edit: I took another look at the recent changes. There was a bug that was causing the initial focus to not trigger the focus border. I didn't pay much attention thinking it would be a simple fix, but seems like opencode pretty puch inlined the entire `PopupMenuItem` component, just to expose the `autofocus` prop from `InkWell`. Which is why the component suddenly exploded in complexity.  I'll try to see if I can make simplify the widget to get it to a more maintainable state.

ashuntu:
> It would mean this widget handles differently from others in Yaru, but that could be more we have work to do in Yaru than this is wrong. @ashuntu not sure if you have any thoughts?

Yeah, if possible it would be nice if we could at least utilize https://github.com/ubuntu/yaru.dart/blob/main/lib/src/widgets/yaru_focus_border.dart here instead of writing a custom wrapper. The idea of `YaruFocusBorder` was for it to be used on things that `yaru.dart` doesn't provide itself, anyway.

Like other discussions have pointed out, this should ultimately live in `yaru.dart`, but if we could at least use `YaruFocusBorder` that would be great (it may not work in this case, I'd be interested in what you find with its behavior here).

packages/security_center/lib/widgets/yaru_popup_menu_item.dart
SoumyaRanjanPatnaik: Fixed.

packages/security_center/lib/widgets/yaru_popup_menu_item.dart
SoumyaRanjanPatnaik: Done.

packages/security_center/lib/widgets/yaru_popup_menu_item.dart
SoumyaRanjanPatnaik: Yes. FIxed.

SoumyaRanjanPatnaik:
I've made the requested changes.

matthew-hagemann (dismissed):
LGTM! Thanks for the changes 🙂

SoumyaRanjanPatnaik:
Thanks. Added a few final changes:
- Set `descendantsAreTraversable: false` on the Focus widget
- Override `buildSemantics` to add a `focused: true` option.

Without these changes, there was an extra focus stop which led to a bad screen reader experience. 

Final behavior with screen  reader:
https://github.com/user-attachments/assets/fa228355-ded0-4d7c-980f-3564f09b66e9

matthew-hagemann (dismissed):
LGTM!
There was a weblate pr that landed that fixed a merge commit that was blocking weblate pr's. You'll just need to rebase on main to grab that fix and we can merge 🙂

SoumyaRanjanPatnaik:
Hi. I've rebased it. CI seems to fail due some fvm related issue.

matthew-hagemann (approved):
Thank you for this! I realize it ballooned from the original ask 🙂 awesome work!

PR #226, opened 2026-05-28
fix(ui): update sidebar icons and copy terminology

juanruitina:
Minor UI fixes for aligning Security Center with designs and language elsewhere in the system.

## Changes

- Fix broken unselected menu icon for Pro page
- Update menu icons for other two pages to match design
- Replace "machine" with "computer" for consistency with Ubuntu Desktop language
- Refer to "sandboxed apps" instead of "strictly confined apps" to match new language (aligned with App Center, Settings app)
- Refer to "Apps" instead of "Applications" to match language in Settings app
- Hide outdated, redundant description line from app permissions page

Note: changes in strings don't substantially change meaning, so we consider this change acceptable at this stage.

## Type of Change

Bug fix (non-breaking)

UDENG-10497


packages/security_center/lib/l10n/app_en.arb
```diff
@@ -74,7 +74,7 @@
     "@microphoneInterfacePageDescription": {},
     "interfacePageTitle": "Manage permissions",
     "@interfacePageTitle": {},
-    "interfacePageDescription": "Strictly confined apps will ask you before accessing files and folders. This feature is experimental and subject to changes.",
+    "interfacePageDescription": "Sandboxed apps will ask you before accessing files and folders. This feature is experimental and subject to changes.",
```
juanruitina: @ashuntu Should we get rid of this string if it's not going to be used anymore?

ashuntu: If a string isn't used anywhere and won't be coming back then it can be removed from the translation files, yes.

juanruitina:
@ashuntu I removed the unused string, this is now ready for review ✌🏻

packages/security_center/lib/routes.dart
```diff
@@ -116,7 +116,7 @@ enum Routes {
       l10n.ubuntuProPageTitle;
 
   static IconData _ubuntuProIcon(bool selected) =>
-      selected ? YaruIcons.ubuntu_logo_simple : YaruIcons.drive_harddisk;
+      selected ? YaruIcons.ubuntu_logo_simple : YaruIcons.ubuntu_logo_simple;
```
ashuntu, line 119:
```suggestion
      YaruIcons.ubuntu_logo_simple;
```

If this icon won't depend on the selected state, then we don't need this conditional.

ashuntu (changes requested):

packages/security_center/lib/routes.dart
```diff
@@ -84,7 +84,7 @@ enum Routes {
         _ => l10n.snapPermissionsPageTitle,
       };
   static IconData _appPermissionsIcon(bool selected) =>
-      selected ? YaruIcons.key_filled : YaruIcons.key;
+      selected ? YaruIcons.app_grid : YaruIcons.app_grid;
```
ashuntu, line 87:
```suggestion
      YaruIcons.app_grid;
```

If this icon won't depend on the selected state, then we don't need this conditional.

ashuntu (changes requested):

ashuntu:
You'll need to format `routes.dart` so CI passes, otherwise this looks good.

ashuntu (approved):
Thanks!

PR #229, opened 2026-06-04
Uniquely handle polkit errors using snapd reported error kinds

### LINKED ISSUE #145: Polkit prompt authentication failure not handled uniquely
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

Currently, if the user is prompted by `polkit` to preform actions and either cancels or provides an inccorect password, we don't catch that error kind response from snapd, and instead display a more generic / incorrect error message.

Image

### Steps to reproduce the behavior

_No response_

### Expected behavior

_No response_

### Ubuntu release

None

### What architecture are you using?

None

### System info

_No response_

### Additional context

_No response_

SoumyaRanjanPatnaik:
## Summary
Improves disk encryption error handling by mapping snapd exceptions to typed TPM FDE operation errors, preserving both the failed operation and snapd error kind. This PR only covers polkit / auth related failures. 

This lets the UI show more specific messages for actions like adding, changing, or removing a PIN/passphrase instead of falling back to generic exception text.

All handled error kinds come from snapd error-kinds definitions defined here:
https://github.com/canonical/snapd/blob/master/client/errors.go#L34-L200

The mapping is intentionally small for now. This can be extended further with more error kinds and for other pages as well by just defining the error-kinds and relevant mappings.

Fixes #145, UDENG-10231

image

## Testing
Added end to end tests for ensuring authentication failures correctly show the error messages. Also tested in a `resolute` VM with TPM FDE setup to verify expected behavior and ensure nothing is broken.

### DIFF
--- a/packages/security_center/lib/disk_encryption/disk_encryption_page.dart
+++ b/packages/security_center/lib/disk_encryption/disk_encryption_page.dart
@@ -89,7 +89,7 @@ class EncryptionPageBody extends ConsumerWidget {
             yaruInfoType: YaruInfoType.danger,
           ),
         _ => YaruInfoBox(
-            title: Text(l10n.diskEncryptionPageError),
+            title: Text(l10n.recoveryKeySomethingWentWrongHeader),
             subtitle: Text(e.toString()),
             yaruInfoType: YaruInfoType.danger,
           ),
@@ -452,9 +452,11 @@ class ChangeAuthDialog extends ConsumerWidget {
     return AlertDialog(
       title: YaruDialogTitleBar(title: Text(title)),
       titlePadding: EdgeInsets.zero,
+      contentPadding: EdgeInsets.zero,
       content: SizedBox(
         width: 460,
         child: ScrollablePage(
+          padding: const EdgeInsets.all(24),
           children: [
             CurrentPassphraseFormField(authMode: authMode),
             PassphraseFormField(authMode: authMode),
@@ -494,9 +496,7 @@ class ChangeAuthDialog extends ConsumerWidget {
               ),
             if (model.dialogState is ChangeAuthDialogStateError)
               YaruInfoBox(
-                title: (model.dialogState as ChangeAuthDialogStateError).fatal
-                    ? Text(l10n.recoveryKeySomethingWentWrongHeader)
-                    : Text(l10n.diskEncryptionPageError),
+                title: Text(l10n.recoveryKeySomethingWentWrongHeader),
                 subtitle: Text(
                   _dialogErrorMessage(
                     (model.dialogState as ChangeAuthDialogStateError).e,
@@ -571,11 +571,13 @@ class ChangeAuthModeDialog extends ConsumerWidget {
     return AlertDialog(
       title: YaruDialogTitleBar(title: Text(title)),
       titlePadding: EdgeInsets.zero,
+      contentPadding: EdgeInsets.zero,
       insetPadding:
           const EdgeInsets.symmetric(horizontal: 40.0, vertical: 16.0),
       content: SizedBox(
         width: 460,
         child: ScrollablePage(
+          padding: const EdgeInsets.all(24),
           children: [
             Text(bodyMainText),
             Text(bodyRecoveryText),
@@ -590,14 +592,9 @@ class ChangeAuthModeDialog extends ConsumerWidget {
                 ),
               ],
             ),
-            if (model.dialogState
-                case ChangeAuthModeDialogStateError(:final e, :final fatal))
+            if (model.dialogState case ChangeAuthModeDialogStateError(:final e))
               YaruInfoBox(
-                title: Text(
-                  fatal
-                      ? l10n.recoveryKeySomethingWentWrongHeader
-                      : l10n.diskEncryptionPageError,
-                ),
+                title: Text(l10n.recoveryKeySomethingWentWrongHeader),
                 subtitle: Text(_dialogErrorMessage(e)),
                 yaruInfoType: YaruInfoType.danger,
               ),
@@ -704,7 +701,7 @@ class _NoneAuthenticationActions extends ConsumerWidget {
         if (hasError) ...[
           const SizedBox(height: 8),
           YaruInfoBox(
-            title: Text(l10n.diskEncryptionPageError),
+            title: Text(l10n.recoveryKeySomethingWentWrongHeader),
             subtitle: Text(
               tpmState.operationError!.causeByMessage(),
             ),
@@ -771,7 +768,7 @@ class _PassphraseAuthenticationActions extends ConsumerWidget {
         if (hasError) ...[
           const SizedBox(height: 8),
           YaruInfoBox(
-            title: Text(l10n.diskEncryptionPageError),
+            title: Text(l10n.recoveryKeySomethingWentWrongHeader),
             subtitle: Text(
               tpmState.operationError!.causeByMessage(),
             ),
@@ -833,7 +830,7 @@ class _PinAuthenticationActions extends ConsumerWidget {
         if (hasError) ...[
           const SizedBox(height: 8),
           YaruInfoBox(
-            title: Text(l10n.diskEncryptionPageError),
+            title: Text(l10n.recoveryKeySomethingWentWrongHeader),
             subtitle: Text(
               tpmState.operationError!.causeByMessage(),
             ),
--- a/packages/security_center/lib/disk_encryption/disk_encryption_page.dart
+++ b/packages/security_center/lib/disk_encryption/disk_encryption_page.dart
@@ -18,6 +18,11 @@ const defaultRecoveryKeyFileName = 'recovery-key.txt';
 
 const yaruProgressSize = 20.0;
 
+String _dialogErrorMessage(Exception e) => switch (e) {
+      final TpmFdeOperationException e => e.causeByMessage(),
+      _ => e.toString(),
+    };
+
 class DiskEncryptionPage extends ConsumerWidget {
   const DiskEncryptionPage({super.key});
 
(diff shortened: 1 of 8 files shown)


SoumyaRanjanPatnaik:
@juanruitina can you please check if the error messages look good, or if any design changes are required. For a summary, here's all of them:
- Error Kind: `auth-cancelled`
  - **Add PIN**: “Authentication was cancelled. Please authenticate to add a PIN.”
  - **Add passphrase**: “Authentication was cancelled. Please authenticate to add a passphrase.”
  - **Change PIN**: “Authentication was cancelled. Please authenticate to change the PIN.”
  - **Change passphrase**: “Authentication was cancelled. Please authenticate to change the passphrase.”
  - **Remove PIN**: “Authentication was cancelled. Please authenticate to remove the PIN.”
  - **Remove passphrase**: “Authentication was cancelled. Please authenticate to remove the passphrase.”

- Error Kind: `auth-login-required`:
  - **Add PIN**: “Please log in to add a PIN.”
  - **Add passphrase**: “Please log in to add a passphrase.”
  - **Change PIN**: “Please log in to change the PIN.”
  - **Change passphrase**: “Please log in to change the passphrase.”
  - **Remove PIN**: “Please log in to remove the PIN.”
  - **Remove passphrase**: “Please log in to remove the passphrase.”
ssphrase.”

matthew-hagemann (dismissed):
Thanks! Impl LGTM and works well 😄

juanruitina:
I have some questions about these errors. Do they need to communicate them to the user?

- If the user cancelled the polkit prompt, they will be aware of it, and that that's a blocker. The polkit prompt tells them authenticating is required, and they would have clicked on "cancel", which is an explicit action on their behalf.
- If the user enters a wrong password, won't the polkit prompt give them that error message?

Is there any other case that warrants providing such an error? I'm wary of adding so many strings when the cause of the error is implicit.

SoumyaRanjanPatnaik:
@matthew-hagemann @juanruitina  I think this is ready for review. Can you guys review the behavior?




https://github.com/user-attachments/assets/751d3398-c0e0-4f66-82a1-2b08e5ffb8d0

SoumyaRanjanPatnaik:
CI failures are during the `setup-fvm` phase and are unrelated to the code changes in this PR.
```
 ==> Tapping leoafarias/fvm
Cloning into '/home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/leoafarias/homebrew-fvm'...
Tapped 44 formulae (58 files, 248.4KB).
Error: Broken pipe
Error: Process completed with exit code 1.
```

packages/security_center/lib/disk_encryption/disk_encryption_page.dart
```diff
@@ -551,6 +551,28 @@ class ChangeAuthModeDialog extends ConsumerWidget {
       _ => '',
     };
 
+    final canSave =
+        model.dialogState is ChangeAuthModeDialogStateInput && notifier.isValid;
+
+    Future<void> handleSubmit() async {
+      final navigator = Navigator.of(context);
+      final operation = notifier.replaceAuthMode();
+      navigator.pop();
```
matthew-hagemann: We are popping too early imo. I'd only pop the dialog if polkit succeeds, so we don't have it re-spawning on failure.

SoumyaRanjanPatnaik: My understanding is, the polkit request is handled by snapd (and not security-center),  so there's no way for us to wait for `polkit` to succeed, without also waiting for the entire `replacePlatformKey` operation to finish.

This is what I had tried initially (by awaiting `notifier.replaceAuthMode()`).  It caused the dialog to stay open for a few more seconds after the polkit authentication, till the auth mode actually changes. The dialog doesn't show any loader, so I felt it right to pop the dialog to let the `diskEncryptionPage` loader to show

matthew-hagemann: The `replacePlatformKey` function is invoking an async snapd api, so it is done in two phases: A sync api call that will block on polkit auth and return us a changeID on success to track, and then we watch that changeID until the operation completes.

In the `packages/security_center/lib/services/snapd_disk_encryption_service.dart` at `replacePlatformKey` you can see them one after the other. We need the changeID back to watch the change, so they feed into one another and we will only get a changeId back from snapd to watch if the polkit prompt was successful. 

So we can add a callback to that function to allow us to know when we had a change ID, which is where we can safely pop the dialog, and avoid waiting for the operation to complete fully:

Hacked together example:

```diff
diff --git a/packages/security_center/lib/disk_encryption/disk_encryption_page.dart b/packages/security_center/lib/disk_encryption/disk_encryption_page.dart
index 32fb509..9f6d013 100755
--- a/packages/security_center/lib/disk_encryption/disk_encryption_page.dart
+++ b/packages/security_center/lib/disk_encryption/disk_encryption_page.dart
@@ -556,21 +556,10 @@ class ChangeAuthModeDialog extends ConsumerWidget {
 
     Future<void> handleSubmit() async {
       final navigator = Navigator.of(context);
-      final operation = notifier.replaceAuthMode();
-      navigator.pop();
-
-      final dialogState = await operation;
-      if (!navigator.mounted) return;
-
-      switch (dialogState) {
-        case ChangeAuthModeDialogStateInput():
-          await showDialog<void>(
-            context: navigator.context,
-            builder: (_) => ChangeAuthModeDialog(authMode: authMode),
-          );
-        case _:
-          break;
-      }
+      // Pop only once polkit has authorized the change, handing off to the disk
+      // encryption page loader. A cancelled or rejected auth never fires
+      // onAuthorized, so the dialog stays open instead of respawning on failure.
+      await notifier.replaceAuthMode(onAuthorized: navigator.pop);
     }
 
     return AlertDialog(
diff --git a/packages/security_center/lib/services/snapd_disk_encryption_service.dart b/packages/security_center/lib/services/snapd_disk_encryption_service.dart
index 6d181d7..47f8f2d 100644
--- a/packages/security_center/lib/services/snapd_disk_encryption_service.dart
+++ b/packages/security_center/lib/services/snapd_disk_encryption_service.dart
@@ -116,6 +116,7 @@ class SnapdDiskEncryptionService implements DiskEncryptionService {
     required AuthMode authMode,
     String? passphrase,
     String? pin,
+    void Function()? onAuthorized,
   }) async {
     final snapdAuthMode = switch (authMode) {
       AuthMode.none => SnapdSystemVolumeAuthMode.none,
@@ -129,6 +130,10 @@ class SnapdDiskEncryptionService implements DiskEncryptionService {
       pin: pin,
     );
 
+    // snapd returns a change id only after polkit auth has passed; the change
+    // itself runs asynchronously below.
+    onAuthorized?.call();
+
     final result =
         await _snapd.watchChange(changeId).firstWhere((change) => change.ready);
     if (result.err != null) {
```

There is a small delay, but the user isn't left in that dialog for multiple seconds.

SoumyaRanjanPatnaik: > we will only get a changeId back from snapd to watch if the polkit prompt was successful.

Ahh... That makes sense. I looked at that part of the code but thought didn't realize polkit auth blocks the async call.

Will make the change. Seems simple enough.

SoumyaRanjanPatnaik:
Made the changes. Have created `fixup` commits for clearer diffs on what changed. Should help you review. Attaching a video of the updated behaviour.

https://github.com/user-attachments/assets/d1a571ed-d80e-464b-ba64-fcb12d848874

juanruitina (commented):
Please use "Something went wrong" for the error panel title

Also, I see some excessive padding in the modal, please make it 24px all around.

Maybe for a separate PR, so leaving as a note:  same as with https://github.com/canonical/desktop-security-center/pull/236, if elevated are always required, we could perhaps move the polkit prompt before opening the modal. That way the user wouldn't be bothering entering a passphrase that they cannot save. It is true that we wouldn't be able to explain what the PIN/passphrase are about.

SoumyaRanjanPatnaik:
Right now, the "Something Went Wrong" error shows only when the error is `fatal`.  Do we want to show "Something Went Wrong" for all the error cases?

packages/security_center/lib/disk_encryption/disk_encryption_providers.dart
```diff
@@ -69,6 +69,97 @@ sealed class SnapdStateException
...
+  authCancelled('auth-cancelled'),
+  authLoginRequired('auth-login-required');
+
+  const SnapdAuthErrorKind(this.snapdKind);
+
+  final String snapdKind;
+
+  static SnapdAuthErrorKind? fromKind(String? kind) => switch (kind) {
+        'auth-cancelled' => SnapdAuthErrorKind.authCancelled,
+        'auth-login-required' => SnapdAuthErrorKind.authLoginRequired,
```
matthew-hagemann, line 85: I don't see `auth-login-required` in https://github.com/canonical/snapd/blob/master/client/errors.go#L34-L200?

matthew-hagemann (changes requested):
Last thing other than needing to rebase:

packages/security_center/lib/disk_encryption/disk_encryption_providers.dart
SoumyaRanjanPatnaik: Yeahh. It was supposed to be `login-required` :sweat_smile:

matthew-hagemann:
Thanks @SoumyaRanjanPatnaik! Can you rebase to resolve the merge conflicts please :slightly_smiling_face:

SoumyaRanjanPatnaik:
Yeah sure.

SoumyaRanjanPatnaik:
Done

SoumyaRanjanPatnaik:
Squashing fixup! commits as well

matthew-hagemann (approved):
Thanks! LGTM!

PR #235, opened 2026-06-17
fix: constrain pages to 575 max width

ashuntu:
From a conversation width @juanruitina, constrains the content of pages to 575 width so they don't look awkward when resized for large screens.

| Before | After |
|--------|-------|
| image | image |

---

UDENG-10512


d-loose (dismissed):

juanruitina:
I made two rather minor changes:

- Padding was _within_ the 575px, so I moved the padding outside the constrained box
- Width should actually be 574px, I just checked with the GTK inspector. Sorry I gave the wrong number!

d-loose (approved):

PR #236, opened 2026-06-17
fix: pre-authorize pro attach with a dummy dbus call

ashuntu:
The current auth flow for Pro can be a bit awkward, especially with the magic attach flow, since you'll still need to enter your password when confirming. This moves the polkit prompt to before the intial dialog is shown initially by calling Pro attach with an empty token.

This is a bit of a hack, but it does make the flow make a little more sense.

---

UDENG-10513

### DIFF
--- a/packages/security_center/lib/services/fake_ubuntu_pro_dbus_service.dart
+++ b/packages/security_center/lib/services/fake_ubuntu_pro_dbus_service.dart
@@ -128,6 +128,9 @@ class FakeUbuntuProManagerService implements UbuntuProManagerService {
   }
 
   @override
+  Future<void> preAuthorize() async {}
+
+  @override
   Future<void> attach(String token) async {
     _stream.add(_data.copyWith(attached: true));
   }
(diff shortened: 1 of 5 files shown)


ashuntu:
@juanruitina right now this will still open the dialog even if auth fails (e.g. you cancel), and you'd be prompted for auth again for real when you actually attach. Should we only allow the dialog to show if you successfully auth, even if we're not actually doing anything with the auth yet at this point in the flow?

juanruitina:
@ashuntu Yep, if the user doesn't successfully auth, the process should end and dialog should not show. It will be implicit that elevated privileges are needed.

ashuntu:
Now that I look into it more, I don't think there's a way we can ensure the authorization succeeds since the way I'm doing this is with a dummy call that will always fail, anyway.

Not sure if there is a better way to do this other than extending the d-bus service itself.

@d-loose Do you have any ideas?

d-loose:
I think this goes against the design principles of polkit, where authentication is directly tied to a specific privileged action, so I'd recommend keeping the current flow.

ashuntu:
> I think this goes against the design principles of polkit, where authentication is directly tied to a specific privileged action, so I'd recommend keeping the current flow.

I'll close this for now then, maybe we can revisit the flow later.

PR #238, opened 2026-07-07
Handle unknown disk encryption status

SoumyaRanjanPatnaik:
## Summary
- handle unknown snapd storage encryption status parse failures without crashing
- map unknown statuses to the existing disk-encryption failure UI and log the parse error
- add a service POC/regression test for actual unexpected enum deserialization

## Testing strategy
Focused service-level regression test that reproduces snapd.dart's actual unknown-enum deserialization failure and verifies it is mapped to a failed disk-encryption status.

Note: broader disk encryption widget tests are currently blocked locally by stale generated code in disk_encryption_providers.freezed.dart referencing TpmFdeOperation/SnapdAuthErrorKind while source uses AuthOperation.


matthew-hagemann (approved):
Thanks! LGTM 🙂

PR #247, opened 2026-09-14
fix: added noto fallback fonts for arm64

ogres:
Added workaround for ARM64 Flutter fonts

UDENG-11438


d-loose (approved):
Very nice! Code changes look good to me. I successfully tested this in a VM now.

snap/snapcraft.yaml
```diff
@@ -66,3 +66,5 @@ parts:
       fvm flutter build linux --release -v
       mkdir -p $CRAFT_PART_INSTALL/bin/
       cp -r build/linux/*/release/bundle/* $CRAFT_PART_INSTALL/bin/
+    stage-packages:
+    - fonts-noto-cjk
```
d-loose, line 70: @ogres on second thought, maybe we can stage those only for arm64 to not unnecessarily inflate the size of the amd64 build? See https://ubuntu.com/docs/snapcraft/9/reference/advanced-grammar/

ogres: Lets try

PR #250, opened 2026-09-18
fix(app-permissions): remove give feedback link

juanruitina:
Remove the "Give feedback" survey link from the App Permissions interfaces page and the corresponding GitHub issue template contact link.

## Rationale

The Maze survey has been running for almost 2 years, with enough evidence collected (>500 responses). A new survey could be run at a later stage.

## Type of Change
Bug fix (non-breaking)


packages/security_center/lib/app_permissions/interfaces_page.dart
d-loose: Instead of making the url an empty string and hiding it from the view, we should just remove this enum member entirely (including the switch branches in the `url` getter and `localize`.

d-loose (changes requested):

packages/security_center/lib/app_permissions/interfaces_page.dart
juanruitina: Done, also removed the translated strings 👍🏻

d-loose (approved):
Thanks, LGTM

